PR #31, opened 2017-11-19
Fix two simple typos

rjc:
Mastodon is a *single* federated social network - albeit decentralised with multiple independent servers.

The 2nd typo is insignificant as the `.list` file can be named anything but I thought you'd like consistency :^)


ihabunek:
Thanks. :)

PR #47, opened 2018-01-31
Add proxy option to config and prepare the requests with it

ckeen:
I need these changes to be able to use toot behind a mandatory proxy. I am not that fluent in python so I am proposing this as a proof of concept. If you want me to change it so it will fit the upstream sources better, please do tell.

Thanks!

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -84,8 +84,12 @@ def post_status(app, user, status, visibility='public', media_ids=None):
     }).json()
 
 
-def timeline_home(app, user):
-    return http.get(app, user, '/api/v1/timelines/home').json()
+def timeline_home(app, user, args):
+    if args.tag:
+        timeline = "tag/%s" % args.tag
+    else:
+        timeline = "home"
+    return http.get(app, user, '/api/v1/timelines/%s' % timeline).json()
 
 
 def get_next_path(headers):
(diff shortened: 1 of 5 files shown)


ihabunek:
Hi! Sorry for ignoring you for a long while.

Turns out that requests has an option to [set the proxies via enironment variables](http://docs.python-requests.org/en/master/user/advanced/#proxies). I made the required changes to support these in 7195a0589092cfb048e2dc141c7e7d1a2b775af9.

Documented here:
https://toot.readthedocs.io/en/latest/advanced.html#using-proxies

Does this work for you?

ckeen:
* Ivan Habunek <notifications@github.com> [181230 11:15]:
> Hi! Sorry for ignoring you for a long while.
> 
> Turns out that requests has an option to [set the proxies via enironment variables](http://docs.python-requests.org/en/master/user/advanced/#proxies). I made the required changes to support these in 7195a0589092cfb048e2dc141c7e7d1a2b775af9.
> 
> Documented here:
> https://toot.readthedocs.io/en/latest/advanced.html#using-proxies
> 
> Does this work for you?

Yes, awesome! Thank you!


-- 
May you be peaceful, may you live in safety, may you be free from
suffering, and may you live with ease.

Issue #52, opened 2018-03-05
Support for local timeline viewer


kakakaya:
As I mentioned in Issue #51 , `toot timeline` 's doc claims to display public timeline, but its implementation is displaying home timeline.

But, anyway there's no way to display local timeline. I'd like to check local timeline using toot command.

ihabunek:
Fixed in #65

PR #53, opened 2018-03-05
Local timeline subcommand

kakakaya:
Implemented local timeline viewer, which I suggested in #52 .

Usage:
`toot timeline --local`


ihabunek:
@kakakaya I merged your changes into a new PR #65 to fix some merge conflicts. I also added support for hashtag and list timelines. So now it will be possible to:

```
toot timeline
toot timeline --public
toot timeline --public --local
toot timeline --tag foo
toot timeline --tag foo --local
toot timeline --list 123
```
This should cover all the options.

Thanks for contributing.

Issue #56, opened 2018-04-05
Fetch instance data using HTTPS, not HTTP


marnanel:
The "login" command always fetches instance data using HTTP:

```
$ toot login --debug -i mastodon.social
Looking up instance info...
DEBUG:toot:>>> GET http://mastodon.social/api/v1/instance
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): mastodon.social
DEBUG:urllib3.connectionpool:http://mastodon.social:80 "GET /api/v1/instance HTTP/1.1" 302 161
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): mastodon.social
DEBUG:urllib3.connectionpool:https://mastodon.social:443 "GET /api/v1/instance HTTP/1.1" 200 None
```

But toot uses HTTPS at no other time, not even later in the login attempt. For example, with a hacked-up server which answers on HTTP:

```
$ toot login --debug -i 127.0.0.1
Looking up instance info...
DEBUG:toot:>>> GET http://127.0.0.1/api/v1/instance
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 127.0.0.1
DEBUG:urllib3.connectionpool:http://127.0.0.1:80 "GET /api/v1/instance HTTP/1.1" 200 245
DEBUG:toot:<<< <Response [200]>
DEBUG:toot:<<< b'{\n  "contact_account": "marnanel",\n  "description": "just a test",\n  "email": "marnanel@thurman.org.uk",\n  "languages": [\n    "en_GB"\n  ],\n  "title": "un_chapeau test",\n  "uri": "http://127.0.0.1",\n  "urls": {},\n  "version": "un_chapeau 0.0.1"
\n}'
Found instance un_chapeau test running Mastodon version un_chapeau 0.0.1
Registering application...
DEBUG:toot:>>> POST https://127.0.0.1/api/v1/apps
DEBUG:toot:>>> DATA:    {'website': 'https://github.com/ihabunek/toot', 'redirect_uris': 'urn:ietf:wg:oauth:2.0:oob', 'client_name': 'toot - a Mastodon CLI client', 'scopes': 'read write follow'}

```

ihabunek:
Fixed via #80 and b06e9ea733674ece381c52674951fbb124722f39. Please check if this works for you.

PR #59, opened 2018-04-19
Add -I / --insecure option, to use HTTP instead of HTTPS.

marnanel:
Add ability to use HTTP instead of HTTPS. This is useful for testing new servers.

Perhaps we should advertise more obviously that this is not recommended in production.

Ref #56 .


ihabunek:
Thanks for contributing and sorry for ignoring you for a while. 

This PR adds `--accept-invalid-certs` option, but does not use it at all.

I merged PR # #80 which also fixes the same issue, so closing this one, sorry.

PR #62, opened 2018-05-05
Fix FreeBSD install

mekanix:
As mentioned https://www.freshports.org/net-im/toot/ the port name is `toot` but the package name is `py36-toot`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ Install the package:
 
 .. code-block::
 
-    pkg install toot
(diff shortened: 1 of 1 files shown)


ihabunek:
Thanks for the PR.
@0mp could you take a look if this seems correct?

0mp:
I confirm! Sorry.

Thanks @mekanix & @ihabunek :smile:

PR #70, opened 2018-07-25
Added an option to reverse the timeline

waweic:
Added the (in my opinion) mission option to reverse the timeline by adding a command and reversing the entries before they get printed out.


ihabunek:
@waweic Looks good. Thanks for the contribution.

PR #72, opened 2018-08-16
Converting `&apos` and `&apos;` to an apostrophe

nloadholtes:
Sometimes toots come through that don't have the semicolon. This should get
both cases.

### DIFF
--- a/toot/output.py
+++ b/toot/output.py
@@ -149,7 +149,7 @@ def print_timeline(items):
         reblogged = item['reblog']['account']['username'] if item['reblog'] else None
 
         soup = BeautifulSoup(content, "html.parser")
-        text = soup.get_text().replace('&apos;', "'")
+        text = re.sub("&apos;*", "'", soup.get_text())
         time = datetime.strptime(item['created_at'], "%Y-%m-%dT%H:%M:%S.%fZ")
 
         return {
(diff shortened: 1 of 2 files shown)


ihabunek:
I haven't encountered this. Does this happen on a specific instance? Sounds a bit strange.

nloadholtes:
Good question. I'll go and see if I can find the account where this was happening. (I didn't post it originally out of respect to privacy of the user whose name caused the issue.)

ihabunek:
@nloadholtes Can you check if the change proposed in #83 helps your use case?

nloadholtes:
I don't think it does. Here's a screenshot of a toot I made this morning with an apostrophe in it, as seen via the curses view:
image

https://octodon.social/@nloadholtes/101347405963562119

Now for the crazy part: I see that on both master and on my branch. 🤦‍♂️ 

I'm starting to think this PR might not be good/correct. If I can't figure this out in 24hrs, I'll close the pr.

dlax:
@nloadholtes the content of this toot as retrieved through the API at https://octodon.social/api/v1/statuses/101347405963562119 contains `&apos;` (with trailing `;`). So the fix from #83 should help I think.

nloadholtes:
@dlax I totally agree it should, I'm just puzzled why I saw that when I tested. I'm 99% sure my brain is still in holiday mode so I've missed something. 😉 

I'm going to go ahead an close this PR, I think the code from #83 is a better choice, if nothing else it has tests and I don't. 😄

PR #77, opened 2018-10-08
Updating the docs with information about and commands for the curses UI

nloadholtes:
Hi, I like the curses UI and wanted to contribute to the docs to let more people know about the commands there.


ihabunek:
Thanks!

PR #78, opened 2018-10-08
Update for the curses ui help menu

nloadholtes:
Adding in the s, A, and B keys. Also adjusted the other lines to make sure everything
lines up in the UI after adding A and B to the list.

(I saw these were missing when I was doing PR #77 so I went ahead and added them in)


ihabunek:
@nloadholtes the A and B shortcuts are actually there to catch arrow keys, and not really intended to be used. :) I'd rather if you removed them, and just keep the s. Thanks.

ihabunek:
Thanks!

PR #80, opened 2018-12-25
Feature proposal: --disable-https flag

SteelPangolin:
`toot` is a great tool for testing modifications to local Mastodon or Pleroma instances, but it needs a way to switch off HTTPS. This change adds a `--disable-https` flag to `toot login` and `toot login_cli`, as well as a `scheme` argument to login-related API methods that don't use `App.base_url`.


ihabunek:
Nice patch. Thanks!

PR #83, opened 2019-01-01
Replace &apos; by "'" before parsing HTML

dlax:
Beautiful will does not parse HTML entities like `&apos;` as we expect
and the previous logic of replacing this *after* HTML parsing occurred
did not produced expected results.

To illustrate this, we change data in "test_timeline" to include a
literal `&apos;` as it sometimes occur in data returned by Mastodon API.
New HTML content is:

    <p>The computer can&apos;t tell you the emotional story [...] </p>

Beautiful will parse this as as:

    <p>The computer can&amp;apost tell you the emotional story [...] </p>

which is not what we expect.

We fix this by replacing `&apos;` *before* HTML parsing by Beautiful.
Since test data in "test_timeline" got updated we also add an extra
assertion checking that part of the content with a literal "'" is
(still) properly rendered.

### DIFF
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -126,7 +126,7 @@ def test_timeline(mock_get, monkeypatch, capsys):
             'username': 'fz'
         },
         'created_at': '2017-04-12T15:53:18.174Z',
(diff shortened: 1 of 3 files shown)


ihabunek:
Thanks! I believe this also solves the issue in #72.

PR #85, opened 2019-01-03
Add local and tag timelines to curses

### LINKED ISSUE #61: Support hashtag timelines
Hi,

The Mastodon API supports returning the timeline of toots with a hashtag, eg: `https://mastodon.social/api/v1/timelines/tag/myhashtag`

Could you add a command to `toot` to show it?

Thanks!

ksunden:
Takes inspiration from the approach taken by timeline command.
Closes #61

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -173,17 +173,17 @@ def get_next_path(headers):
         return "?".join([parsed.path, parsed.query])
 
 
-def _timeline_generator(app, user, path, limit=20):
+def _timeline_generator(app, user, path, params=None):
     while path:
-        response = http.get(app, user, path)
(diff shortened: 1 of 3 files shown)


dlax:
Tested, works fine. The implementation looks good to me as well.

ksunden:
I thought about adding "list" as well, but I don't actually use that feature (also the short options `-l` and `-i` are both taken on `curses`, perhaps `-L` would be good?)

ihabunek:
Looks good, thanks for the patch. If you wanted to implement a list, that would be appreciated. `-L` sounds good, or perhaps we can just not have a short option for that?

PR #88, opened 2019-01-24
Add various features around "reblog"

dlax:
* add a "reblogged_by" command showing accounts that reblogged a status
* display whether the user reblogged a status in curses ui
* add an action to reblog/unreblog a status in curses ui


dlax:
If needed, I can split this into separate PR, just let me know.

ihabunek:
Thanks! Merged manually with some minor tweaks.

The naming of actions if sometimes weird. API uses "reblog" while UI uses "boost". I changed it to "boost" for now to be consistent with the mastodon UI.

PR #96, opened 2019-02-17
Handle notifications

dlax:
Adding support for handling notifications using a "notifications" command.

By default, it will list notifications for current user, most recent first. The list will also contain any status involved in the notification (e.g. for "reblog" or "favourite" notification type), perhaps this is a bit too verbose?

With a `--clear` flag, the command will clear user's notification from the server.

### DIFF
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -545,6 +545,20 @@ def test_notifications(mock_get, capsys):
         "",
     ])
 
+@mock.patch('toot.http.get')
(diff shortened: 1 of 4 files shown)


ihabunek:
Thanks! Merged manually.

I moved the logic for printing to `output.py`. Also removed sorting of notifications, since to me it seems they are already sorted by date descending. Correct me if I'm wrong.

PR #99, opened 2019-03-29
Compose / reply functionality in curses interface

Skehmatics:
Add a reply (r) and compose (c) action for the curses interface complete with adding a cw (^s)
Uses the regular draw_lines to draw the textbox, EOT or Ctrl+Enter (not well supported on all terminals) to confirm / post, and updates the timeline with the new status.

The use of draw_lines means that linebreaks due to width work, but strip whitespace down to one character on the display despite them still being included in the text to be sent, which is a tad confusing at first. I'm also working on an [improvement ](https://github.com/Skehmatics/toot/tree/feature/preserve_whitespace) to the fit_text / wc_wrap function that treats whitespace a little more intelligently throughout the interface, but there's a lot of tiny issues I want to work out on that first so it's not included here.

Let me know if you have any concerns or want this all squashed to one commit :)

### DIFF
--- a/toot/ui/app.py
+++ b/toot/ui/app.py
@@ -317,38 +317,86 @@ class HelpModal(Modal):
         ]
 
 
-class ComposeModal(Modal):
-    def __init__(self, stdscr):
-        self.height, self.width, y, x = self.get_size_pos(stdscr)
+class TextModal(Modal):
+    def __init__(self, stdscr, title, footer=None, maxsize=(None, None)):
+        self.pad_y = 2
+        self.pad_x = 2
+        self.title = title
+        self.footer = footer
+        if self.footer:
+            self.pad_y += 1
+        self.height, self.width, y, x = self.get_size_pos(stdscr, maxsize)
 
         self.window = curses.newwin(self.height, self.width, y, x)
-        self.text_window = self.window.derwin(self.height - 6, self.width - 5, 3, 2)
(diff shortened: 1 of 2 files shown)


ihabunek:
Nice work! Composing and replying seems to work well.

Merged manually. I changed the content warning shortcut from ^S to ^W since ^S will freeze most terminals.

ihabunek:
@Skehmatics what was the issue with `texpad.TextBox` that you implemented your own model for text entry?

Skehmatics:
@ihabunek as touched on in 6e53dc3's commit message, `textpad.TextBox`'s `edit()` and `gather()` functions leave the newlines it adds for screen wrapping in the string, making [some janky looking toots](https://is.nota.live/@skehmatics/101661925423868292). I felt that correctly removing those while keeping intentional newlines would be more difficult to do right.

ihabunek:
@Skehmatics Right, makese sense. With the growth of complexity of the curses interface, I'm looking at [urwid](http://urwid.org/) for common components, but frankly it confuses the hell out of me every time I try to put something together. I haven't really found other libs like [npyscreen](https://npyscreen.readthedocs.io/introduction.html) to be much better.

Skehmatics:
[asciimatics.widgets](https://asciimatics.readthedocs.io/en/latest/widgets.html) worked quite well for me on another project, but it would be quite a port since its actually not based on curses at all (which, on the upside, means windows compatibility), and I'm not certain it would improve the situation all that much.

PR #101, opened 2019-04-25
Fix various resize issues

Skehmatics:
+ Fix resizes not working when modals are open (help modal blocks event, compose modal updates but does not resize) via `resize_callback` arg on Modal. It's a little jittery, but functional.
+ Fix minor layering issue with footer


ihabunek:
Thanks. A bit messy but works, I'll work on the aesthetics later. :)

PR #102, opened 2019-04-25
Fix #100

### LINKED ISSUE #100: Disappearing status list
To reproduce:

* open toot curses

image

* press "c" to compose a new toot, status list disappears:

image

The issue does not occur if the user moves down the status list by pressing `j` before composing a toot.

Skehmatics:
Using self.panel.hide() without calling
curses.panel.update_panels() before next refresh can really mess
stuff up. As far as I can tell, this was the real cause of #100 .

This changes that call to the Modal's hide method instead, which
will keep this sort of issue from coming up again (hopefully)


ihabunek:
Thanks, nice catch. I merged it locally. In the future, please use `git commit --amend` to fix existing commits.

PR #105, opened 2019-06-18
Bind <c-r> to show status context in curses UI

### LINKED ISSUE #104: Ability to view thread from curses view
I would like to be able to press the side-arrow (right) to view a thread of replies.

dlax:
This binds <enter> (resp. <escape>) to switch to the thread view from selected toot in the curses UI.

It's a bit hackish at the moment, but if there's interest in this feature I can rework things. Feedback welcome in any cases.

Would fix #104.

### DIFF
--- a/toot/ui/app.py
+++ b/toot/ui/app.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
+
+from copy import copy
 import os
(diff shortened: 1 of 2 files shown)


ihabunek:
Status context has been implemented in #108 in the new TUI.

PR #110, opened 2019-08-31
Indicate if status is a reply in timeline of TUI

dlax:


ihabunek:
Thanks! Merged manually. I just added `cyan` color to the `attr_map` so that it displays with a green background when selected.

https://github.com/ihabunek/toot/commit/19aa56e1fe44bd13ed98c393721c319d98233d4e#diff-e621a1224adf73b4047b7d548da9931fR59

dlax:
> I just added cyan color to the attr_map so that it displays with a green background when selected.

thanks, that's indeed nicer!

PR #112, opened 2019-09-02
Make "view" action work for reblogged statuses

dlax:
When the status is a reblog, the URL of the original toot should be used
as the one of the reblog is null.

We add a "url" attribute to Status class storing computed URL.


ihabunek:
Rebased & merged, thanks!

PR #113, opened 2019-09-03
Remove double process_response function call, since it's called in anon_post already

l-n-s:


ihabunek:
Thanks!
Merged manually: 616483d78a02dfdd7586f02e7094a148f1a50f1b

PR #120, opened 2019-09-06
Escape closes timeline

RasmusLindroth:
The help message says that you can use Esc and Q to go back and close overlays. So it seems logical that you should be able to use Esc to go back when viewing a thread.


ihabunek:
@RasmusLindroth Thanks! I merged it manually with a little tweak: i don't want ESC to quit the app if the main timeline is active. Check out 3a562438783e12e8ea7696b45f0f778e4604eeb5.

PR #125, opened 2019-09-13
Try to use media's text_url if present

dlax:
This field is sometimes provides and gives a shortened URL which is
easier to click on in a terminal.


ihabunek:
Thanks!

PR #126, opened 2019-09-13
Include mentions and replied-to account in compose text

dlax:
We add a "mentions" attribute to Status. Then when composing a reply, we
fill the edit text of the compose box with the account name of status
being replied to and possibly include mentions at the bottom of the edit
text. Initial cursor position is set after replied account name.


ihabunek:
Thanks!

PR #127, opened 2019-09-13
Set urwid global encoding to UTF-8

bearzk:
so that many characters can be displayed correctly.

before applying this change:
image

after applying this change:
image


ihabunek:
I don't have  this problem locally. This is probably due to your locale settings, urwid guesses the encoding based on them.

On the other hand, assuming UTF8 is also ok, so merging. Thanks for the pull!

PR #128, opened 2019-09-16
Improve handling of reblogs in TUI 

dlax:
A couple of changes to improve handling of reblogs in the urwid TUI, mostly to retrieve data from the original status (the one that's reblogged) instead of from the reblog.

### DIFF
--- a/toot/tui/app.py
+++ b/toot/tui/app.py
@@ -335,7 +335,7 @@ class TUI(urwid.Frame):
         promise.add_done_callback(lambda *args: self.close_overlay())
 
     def show_media(self, status):
-        urls = [m["url"] for m in status.data["media_attachments"]]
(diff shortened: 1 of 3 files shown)


ihabunek:
Thanks, I'll have a look at both PRs soon. It's been a busy time for me lately.

dlax:
No problem, take your time!

ihabunek:
@dlax This seems add more and more code to handle the same problem for various fields. What do you think about this approach?
https://github.com/ihabunek/toot/pull/130

dlax:
> What do you think about this approach? https://github.com/ihabunek/toot/pull/130

That's indeed cleaner and more explicit as it does not shadow
information of the reblogging status.

ihabunek:
OK. Going forward with that version. Thanks for this though.

PR #132, opened 2019-10-01
Fall back to "username" when "display_name" is unset

dlax:
We add a "username" field to Author entity, this is then used when
displaying the "reblogged by" information when respective account has no
display name.


dlax:
forgotten PR?

ihabunek:
Yes. :) Thanks.

PR #153, opened 2020-01-26
Fix VISIBILITY_OPTIONS

### LINKED ISSUE #152: The explanations of "Unlisted" and "Private" are reversed.
The interface says:
> Private - Do not post to public timelines
> Unlisted - Post to followers only
However posts with the visibility "Private", later viewed on the web, show up with a visibility status as "Unlisted", and vice versa.

finnoleary:
Reorder Visibility Options to be from "most visible" to "least visible", and swap the definitions of "private" and "unlisted" to match the actual visibility that results from those actions.

Fixes #152


ihabunek:
Merged by hand.

PR #154, opened 2020-01-26
Add an action to display status links and open them

dlax:
We add a new [L]inks action that opens an overlay window with links
found in the content of selected status. Links are selectable and upon
click/enter we open the web browser at link's URL.


ihabunek:
Thanks!

PR #155, opened 2020-01-26
Document the [M]edia action

dlax:


ihabunek:
The media action is really not very well implemented. I'd prefer to leave it undocumented until this is done in a better way. And that is probably blocked until some sort of user configuration is implemented where they could chose the image viewer. I'll leave this PR open for now.

dlax:
Ok, makes sense!

dj-bauer:
Why don't we use xdg-open to open the image?

miKa-pyon:
> Why don't we use xdg-open to open the image?

I would like to set a program just for toot specifically, including command-line arguments for that program.

Personally I have replaced my xdg-open with a dummy that does nothing.

ihabunek:
Thanks! I can finally merge this since configuring the media viewer was implemented recently.

```
λ toot tui -h
Usage: toot tui [OPTIONS]

  Launches the toot terminal user interface

Options:
  -r, --relative-datetimes  Show relative datetimes in status list
  -m, --media-viewer TEXT   Program to invoke with media URLs to display the
                            media files, such as 'feh'
  -c, --colors TEXT         Number of colors to use, one of 1, 16, 88, 256,
                            16777216, 24bit, defaults to 16 if using --color,
                            and 1 if using --no-color.
  -h, --help                Show this message and exit.
```

PR #163, opened 2020-04-15
Implement save (pin) of timelines in TUI

dlax:
These changes implement the ability to save (pin) a timeline selected from the goto menu through the [P] key (also used to un-pin a saved timeline). The information is saved to the configuration at the moment. If there is a way to store on server side, I can also implement this. Just wanted to gather a bit of feedback on the idea first.


ihabunek:
Rebased and merged manually. Thanks for the work!

PR #171, opened 2020-06-10
Add homebrew instructions to install docs

BTBTravis:
Hello @ihabunek big fan of this project! Recently added it to homebrew so mac users have an easy install method. I see [repology.org](https://repology.org/project/toot/versions) has already picked up the change. 

Homebrew formula: https://github.com/Homebrew/homebrew-core/blob/master/Formula/toot.rb

I've subscribed to notifications on this repo and I'll update homebrew when new versions come.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -103,6 +103,17 @@ Install from PyPI using pip, preferably into a virtual environment.
 
     pip install --user toot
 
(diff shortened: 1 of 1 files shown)


ihabunek:
Thanks! Glad to see there are mac users too. :)

Issue #169, opened 2020-05-23
Update copyright year


fturco:
The README still says:

> Copyright 2018 Ivan Habunek <ivan@habunek.com>

Since we're in 2020 now, I suggest updating it.

BTBTravis:
Created PR to update it. Took about as much time as to write your comment. Welcome to contribute @fturco .

ihabunek:
There, fixed. :)

PR #172, opened 2020-06-10
copyright year 2018 --> 2020

BTBTravis:
solves #169


ihabunek:
@BTBTravis Thanks, but I'll skip this one. I don't think the copyright is void if i don't have the current year in the license file. I also don't want to `year++`.

PR #176, opened 2020-07-29
TUI: Use original toot for replies and thread

justjosias:
Previously if you tried to reply to a boosted toot, the username automatically put in the reply box was the username of the one who boosted it, rather than the one who originally posted it. 

Also, when you opened a boosted thread, it would go to the boosted thread, rather than the original thread.

This fixes both of those issues by using the original
toot for both replies and threads.


justjosias:
Note: This may also be useful elsewhere, such as in favorites and boosts, but I found this most pressing, and the changes worked immediately.

PR #189, opened 2021-06-16
WIP: Prevent toot timeline to break when piped to extenal programs

stacyharper:
Toot timeline try to ask the user to continue in the next page. This
dont works if the command is piped to any other software.

We stop the query in the first loop when not in a tty context. The
command should be called with the expected item count in this case.

### DIFF
--- a/toot/commands.py
+++ b/toot/commands.py
@@ -52,7 +52,7 @@ def timeline(app, user, args):
 
         print_timeline(items)
 
-        if args.once:
+        if args.once or sys.stdin.isatty():
             break
 
         char = input("\nContinue? [Y/n] ")


Eluminae:
It is very interesting but piping to less still give a tty context. I guess that because we should be capable to send things to stdin. I tried a grep pipe and I still was able to Yes/No. So I think that toot cannot predict what to do and that we should just use the --once arg when needed. 

Tells me guys if you agree with me :D

ihabunek:
Hey, thanks for the PR. You were close to the solution, you should have tested `sys.stdout.isatty()` instead of `sys.stdin.isatty()`. I fixed up the commit and merged: 6115cea43e1bc603b96b2ce3d18918117f7b3fab

PR #191, opened 2021-08-26
WIP alt text for uploaded media

ansuz:
Hi!

I recently set up my own Mastodon instance and wanted to experiment with automating some posts. I looked around and `toot` seemed to be the simplest way to get started. I found it really easy to use, so thanks for this awesome project!

I read through [toot's documentation](https://toot.readthedocs.io/en/latest/usage.html) and its code, but didn't see any way to add alt-text to uploaded media. I also read some of [the Mastodon API documentation](https://docs.joinmastodon.org/methods/statuses/media/) and found that it accepts a "description" attribute. This PR is my incomplete attempt at adding support for it.

I was able to add a `--description` command and verify that it was interpreted correctly by the command line argument parser. After that I sort of fumbled through the code and added named `description` arguments anywhere it looked like it might help.

After a little experimentation I managed to get _something_ to go through to my Mastodon server, however, instead of the supplied description (a string, which I confirmed with a `print` call) what seems to be an internal value from the `Requests` library (`"#<ActionDispatch::Http::UploadedFile:0x00007fa2aff698d0>"`) is displayed in the Mastodon web client.  This can be seen in [a post that was uploaded in this manner](https://social.cryptography.dog/@bot/106823134553346363) with `toot post --description "a cute fluffy animal" --media cute-animal.png`.

I am not an expert with Python, so I'm hoping this is from a trivial mistake on my part. I'm stuck as to how to investigate further, so I'm just sharing my current progress in the hope that a maintainer or another contributor finds it useful.

Thanks again for this very fun and useful tool!

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -92,6 +92,7 @@ def post_status(
     spoiler_text=None,
     in_reply_to_id=None,
     language=None,
+    description=None,
 ):
     """
     Posts a new status.
@@ -110,6 +111,7 @@ def post_status(
         'spoiler_text': spoiler_text,
         'in_reply_to_id': in_reply_to_id,
         'language': language,
+        'description': description,
     }, headers=headers).json()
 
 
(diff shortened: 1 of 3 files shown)


toot/api.py
```diff
@@ -215,9 +217,21 @@ def anon_tag_timeline_generator(instance, hashtag, local=False, limit=20):
...
+    # the following for the 'alt' and 'title' attributes:
+    # "#<ActionDispatch::Http::UploadedFile:0x00007fa2aff698d0>"
+
+    # I also tried supplying http.post with a named data argument
+    # data={ 'description': description }
+    # but that seems to have no effect
     return http.post(app, user, '/api/v1/media', files={
-        'file': file
+        'file': file,
+        'description': description,
```
ihabunek, line 234: You were on the right track with `data={ 'description': description }`. I tried it out now and it does work, not sure why it didn't for you.

toot/console.py
```diff
@@ -297,6 +297,10 @@ def editor(value):
                 "help": "path to the media file to attach (specify multiple "
                         "times to attach up to 4 files)"
             }),
+            (["-d", "--description"], {
+                "type": str,
+                "help": 'A plain-text description of the media, for accessibility purposes.'
+            }),
```
ihabunek, line 303: One issue: since toot supports multiple `--media` options, it also needs to support multiple `--description` options.

ihabunek (commented):
Hey, thanks for the work. You were close to getting this to work, see comments.

If you don't mind, I'll finish it up and merge.

ihabunek:
Here's the cleaned up commit: https://github.com/ihabunek/toot/commit/ec097886d85acfcb34ef5ef1d5257a46e9f6e6b6.

Thanks again!

ansuz:
Thanks for finishing it up! I've already updated to the newer version and started using it.

Issue #166, opened 2020-04-25
Scroll right-hand pane in TUI


d5xtgr:
On instances where the allowed number of characters per toot is large, or on posts with previews of included links, the content does not always fit in the provided space in the curses interface.  While it is possible to scroll the left-hand pane by means of the arrow keys, there does not appear to be any way to scroll the right-hand pane to view the overflowed content.

ihabunek:
This is something I've been putting off for a while because it means i need to handle focus of left vs right pane. I'll try to work on it soonish.

XorOwl:
Is it not as simple as say, binding +/- on the keypad to make the message panel scroll? No need for added focus and such, just scroll it so its visible

PR #193, opened 2021-10-09
[Z]oom command that opens status in scrollable popup window

PeterFidelman:
Added a [Z]oom command that opens the current status in a scrollable popup window similar to the window used by the view source command.

This is one way of getting the message scrolling requested in #166.

Here's what it looks like:

toot-zoom


ihabunek:
Merged. Thanks. Sorry for the slight delay. ^^;

PR #194, opened 2021-10-09
Show media attachments in the "L" list of links

PeterFidelman:
Current behavior: the "L" command shows list of URLs found in the post and lets you choose a link to open in your web browser.

New behavior: The list now includes media attachments too.  Attachments are labeled with their alt-text if present, otherwise the URL is used.

toot-media

This lets you pop open media attachments without resorting to "V"iewing the whole post in a browser.

### DIFF
--- a/toot/tui/app.py
+++ b/toot/tui/app.py
@@ -325,6 +325,11 @@ class TUI(urwid.Frame):
 
     def show_links(self, status):
         links = parse_content_links(status.data["content"]) if status else []
+        post_attachments = status.data["media_attachments"] or []
+        reblog_attachments = (status.data["reblog"]["media_attachments"] if status.data["reblog"] else None) or []
+        for a in post_attachments + reblog_attachments:
+            url = a["remote_url"] or a["url"]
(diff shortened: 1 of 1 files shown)


PeterFidelman:
Note: Media attachments in boosted posts still do not show up in the list.  I have a patch in testing that may fix that, but I want to put some more mileage on it before I submit it.

PeterFidelman:
An updated version of the patch is now available, that correctly handles the media attachments of boosted posts.

FiskFan1999:
Nice job on this PR, been using the TUI on your branch. Hope this will get merged soon.

My only question regarding this is if it is at all possible to have nested reblogs. From viewing the source, there is a "reblog" key which `= null` in the post which itself is within `data["reblog"]`. Wonder if it would be interesting to go through the reblogs recursively...But most likely this is completely impossible to occur in real life.

dlax:
In case you didn't know it (because it's hidden IIRC), there is a `m` key binding that shows media attachments; but it only works for images it seems.

ihabunek:
Thanks!

PR #200, opened 2022-04-14
Add warning for sensitive (nsfw) posts

FiskFan1999:
Note that this warning is ignored if data["sensitive"] == True but data["spoiler_text"] is non-empty, which is possible when the user writes a post with a spoiler. Figured that would be warning enough.

Of course, am open to formatting suggestions as well.


ihabunek:
Sorry this never got merged, check out [tooi](https://codeberg.org/ihabunek/tooi) if you'd like. Hopefully it does the right thing. :)

PR #212, opened 2022-11-06
Local time in toot timeline

mmmmmmbeer:
Currently, `toot tui` displays local times, but `toot timeline` displays the time in UTC. This change adjusts the date in `toot timeline` to the local (system) timezone, so the timezone is now consistent between the two interfaces.


ihabunek:
Thanks!

PR #213, opened 2022-11-09
use /v2 api for account search

kaja47:
Some servers (namely gotosocial) do not support /api/v1/accounts/search. This patch adds fallback to v2 search when 404 is encountered.

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -238,17 +238,11 @@ def search(app, user, query, resolve):
 
 
 def search_accounts(app, user, query):
-    try:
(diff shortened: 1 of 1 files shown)


ihabunek:
We should just always use the v2 endpoint. It has been out since summer 2018, and v1 is deprecated.

So could you remove the try/catch and just always use v2? You can also add `"type": "accounts"` to the params so it only searches accounts.

Thanks.

ihabunek:
Thanks!

PR #216, opened 2022-11-09
Added install instructions for fedora linux.

nrekretep:
Installation of toot on fedora is simple and obvious but should be still mentioned in the docs for people newly starting with fedora.


ihabunek:
Thanks

PR #220, opened 2022-11-11
Add a suffix to temporary files when composing toot in an editor

### LINKED ISSUE #219: Set an explicit suffix on temporary filenames for toots
When you write a toot using `toot post -e`, the editor opens a temporary file named something like `/tmp/tmp1234567`. Because of the generic filename it's not possible to apply any automatic settings (syntax, wrapping, spell checking, etc) to the file.

Include an explicit prefix or suffix in the name so that toots are identifiable for the purpose of customizing editor settings.

E.g.:

```
with tempfile.NamedTemporaryFile(suffix='.toot') as fd:
  ...
```

larsks:
This enable suffix-based configuration in your editor when editing toots.

Closes #219

### DIFF
--- a/toot/utils.py
+++ b/toot/utils.py
@@ -102,7 +102,7 @@ def editor_input(editor, initial_text):
     """Lets user input text using an editor."""
     initial_text = (initial_text or "") + EDITOR_INPUT_INSTRUCTIONS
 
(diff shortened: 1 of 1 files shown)


ihabunek:
Thanks!

Issue #231, opened 2022-11-19
toot post -e vim not working


vodrazka:
hi,

i tried using vim for posting but i get this error:
`You must specify either text or media to post.`

(i fill the file with text above the specifed line)

I get the same error with neovim (nvim command).

It works with nano. It's probably a configuration issue, but i can't figure out whats the reason.
The temp file *.toot is created and accessible (writable)...

sharils:
@vodrazka regardless of whether #244 is accepted or not, you can add `set backupcopy=yes` to make it work now.

ihabunek:
I'm unable to reproduce this issue. I'm using vim for posting from toot regularly, from Debian and Fedora.

My vim settings have `backupcopy=auto` and `backup=nobackup`. Do you have backup enabled in settings?

vodrazka:
I have backups turned off. Enabling them fixes this.

My init.lua (neovim) has this now:
```
-- toot
vim.opt.backupcopy = 'yes'
```

sharils:
Hi @ihabunek - I’m on `VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Nov 25 2022 21:13:47)`. `:set backup?` is `nobackup`. `:set backupcopy?` is `backupcopy=yes`. I can replicate if `:set backupcopy=auto`.

PR #244, opened 2022-12-03
Add vim modeline to EDITOR_INPUT_INSTRUCTIONS #231

sharils:
To solve #231, there are three options I can think of:

1. Open the temp file again after calling `subprocess.run`, which adds overhead to all other text editors
2. Ask individual vim users to add `set backupcopy=yes` to their vimrc file, which requires every vim users to set backupcopy which isn't default
3. Change vim setting by adding modeline to the EDITOR_INPUT_INSTRUCTIONS, which adds couple more characters to EDITOR_INPUT_INSTRUCTIONS

It seems that option 3 is a middle ground between overhead and user experience, thus this PR.

P.s. Instead of rename the file and write a new one, `backupcopy=yes` forces vim to make a copy of the file and overwrite the original one.

### DIFF
--- a/toot/utils.py
+++ b/toot/utils.py
@@ -99,6 +99,7 @@ EDITOR_INPUT_INSTRUCTIONS = f"""
 Do not modify or remove the line above.
 Enter your toot above it.
 Everything below it will be ignored.
+vim:backupcopy=yes
 """
 
 


sharils:
@ihabunek Would you mind reviewing this? Thanks.

PR #257, opened 2022-12-10
Added bookmark status feature to TUI

danschwarz:
New bookmark timeline
Bookmark/unbookmark keys
Bookmarked statuses are not flagged in timeline (couldn't get a nice Unicode icon that would fit in well without alignment issues?)  but they are flagged in the status detail with a red "bookmark" icon.

Oh, there's one unrelated fix in here, to ExceptionStackTrace overlay - updated for Python 3.10 compatibility.

Screenshot 2022-12-09 205003


ihabunek:
Thanks! Merged manually. I had to change the shortcut to "O" because "K" is used to scroll up (vim style), and "M" is used to show media. Not most intuitive, but what can you do.

danschwarz:
Vim shortcuts.. never even though t to look for them as I use emacs :/ I’ll
look out for them now.

PR #259, opened 2022-12-13
Fix for issue 258 "Sensitive" status display

danschwarz:
Simple indentation change, now hides cards and polls and media attachments


ihabunek:
Thanks!

Issue #264, opened 2022-12-21
Mastodon 4 lets you follow tags. Toot should provide this functionality


danschwarz:
... pull request incoming...

PR #265, opened 2022-12-21
Console and api support for following hashtags

danschwarz:
Mastodon 4 allows you to follow hashtags. This pull request implements basic console and api level support for the feature.

```
toot follow_tag tagname
toot unfollow_tag tagname
toot followed_tags

```
The commands have slightly awkward names as they can be confused with the account follow / unfollow commands.  If you can think of better names, it might be worth switching.

* note: this pull request is a subset of pull request #266 so please accept one or the other, but there is no need to accept both.

https://github.com/ihabunek/toot/issues/264

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -23,6 +23,11 @@ def _status_action(app, user, status_id, action):
 
     return http.post(app, user, url).json()
 
+def _tag_action(app, user, tag_name, action):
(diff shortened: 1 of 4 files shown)


ihabunek:
Thanks! I merged this manually and made some tweaks, most notably, I renamed the commands to: 

* `toot tags_followed`
* `toot tags_follow`
* `toot tags_unfollow`

This seemed a bit more readable, what do you think?

PR #270, opened 2022-12-25
Scrolling status details and inline ANSI graphics previews

danschwarz:
This pull request adds scrolling status details (right-arrow in timeline shifts focus to the status detail window, then up/down arrows allow you to scroll.  left-arrow gets you back to the timeline.)  It also adds ANSI graphics previews for card images.  Images are asychronously loaded.   This could be extended to add previews for media attachments (.png, .jpg, etc.) but there are some threading issues to work out before that can be done properly.
Screenshot 2022-12-24 194323

Note that this does add two new library dependencies; Pillow (PIL) and ansicolors.


danschwarz:
Missing files in this pull request. Resubmitting.

Issue #271, opened 2022-12-25
TUI doesn't refresh after launching and exiting Elinks


danschwarz:
If you view a link using a console webbrowser such as Elinks, when you are done viewing it and you quit to go back to the TUI, you are left with a blank screen and have to manually refresh (ctrl-L) to see the TUI again. 

Pull request incoming for this.

ihabunek:
Fixed in 0.33.1

PR #273, opened 2022-12-25
Screen refresh after console-web-browser invocation and exit

danschwarz:
This checks to see if you're launching an in-console webbrowser and if so, after quitting the browser, it forces a screen refresh so you see the TUI again.  Fix for #271

### DIFF
--- a/toot/tui/app.py
+++ b/toot/tui/app.py
@@ -208,6 +208,7 @@ class TUI(urwid.Frame):
         urwid.connect_signal(timeline, "links", _links)
         urwid.connect_signal(timeline, "zoom", _zoom)
         urwid.connect_signal(timeline, "translate", self.async_translate)
+        urwid.connect_signal(timeline, "clear-screen", self.loop.screen.clear)
 
     def build_timeline(self, name, statuses, local):
         def _close(*args):
@@ -345,6 +346,9 @@ class TUI(urwid.Frame):
             title="Status source",
         )
 
+    def _clear_screen(self,  widget):
+            self.loop.screen.clear()
+
     def show_links(self, status):
         links = parse_content_links(status.data["content"]) if status else []
         post_attachments = status.data["media_attachments"] or []
@@ -353,8 +357,10 @@ class TUI(urwid.Frame):
             url = a["remote_url"] or a["url"]
             links.append((url, a["description"] if a["description"] else url))
         if links:
+            sl_widget=StatusLinks(links)
+            urwid.connect_signal(sl_widget, "clear-screen", self._clear_screen)
             self.open_overlay(
-                widget=StatusLinks(links),
+                widget=sl_widget,
                 title="Status links",
                 options={"height": len(links) + 2},
             )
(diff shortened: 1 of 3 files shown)


toot/tui/app.py
```diff
@@ -208,6 +211,7 @@ def _zoom(timeline, status_details):
         urwid.connect_signal(timeline, "links", _links)
         urwid.connect_signal(timeline, "zoom", _zoom)
         urwid.connect_signal(timeline, "translate", self.async_translate)
+        urwid.connect_signal(timeline, "clear-screen", _clear_screen)
```
ihabunek, line 214: You can use `self._clear_screen` here instead of defining it above.

toot/tui/app.py
```diff
@@ -345,6 +349,9 @@ def show_status_source(self, status):
             title="Status source",
         )
 
+    def _clear_screen(self,  widget):
+            self.loop.screen.clear()
```
ihabunek, line 353: Double indent here.

toot/tui/overlays.py
```diff
@@ -30,17 +30,27 @@ def __init__(self, status_details):
...
         super().__init__(walker)
 
+    def browse(self, url):
+        webbrowser.open(url)
+        wb = webbrowser.get()
+        if wb.name == 'www-browser':
+            # 'www-browser' implies a generic
+            # console browser; once it exits
+            # we need to refresh the tui screen
+            self._emit("clear-screen")
```
ihabunek, line 52: In my testing, I don't get `www-browser` when I configure  elinks as browser. 

```
λ BROWSER=elinks python                                                     
Python 3.10.8 (main, Nov 14 2022, 00:00:00) [GCC 12.2.1 20220819 (Red Hat 12.2.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import webbrowser
>>> webbrowser.get().name
'elinks'
```

I'm guessing you're using the [www-browser ubuntu package](https://packages.ubuntu.com/bionic/www-browser). 

Does it hurt if we clear the screen after exiting any browser, even if it's graphical?

ihabunek (commented):
Thanks! This is useful. I have some comments.

Additionally, when you have some changes to make to code you already changed, like the "whitespace fixes" commit in this pull request, you should amend your commit with the changes instead of creating a new commit.

There's a very good guide on this here: https://git-rebase.io/

That way we don't get "fixed whitespace" commits in master when merging a PR.

Thanks!

toot/tui/overlays.py
danschwarz: Correct, I am using www-browser in ubuntu.  I switched to Firefox for testing purposes and I can confirm that refreshing the screen after exiting any browser has no negative effect. I didn't even see a screen flicker in the console.  So I'll remove the check and execute the refresh every time.

toot/tui/app.py
danschwarz: Fixed

toot/tui/overlays.py
danschwarz: Fixed

danschwarz:
used rebase and force pushed the updates to my browserfix branch so everything should look clean in the pull request (I hope...)

ihabunek:
Excellent. Thanks.

PR #277, opened 2022-12-30
Scrolling status details window and ANSI graphics preview of card images

danschwarz:
Fifth time's a charm


danschwarz:
The latest changes to image5 are significant enough to deserve an extra git log entry imho;  did some research on Python threads, dicts, and after testing I believe the async loading of media attachments is safe to enable. this latest commit does so. 
image

danschwarz:
I brought this pull request up to date with 0.33.1 and made all necessary changes so the code passes the test suite (on my local machine). Unfortunately it's not passing the tests on github because of a weird dependency issue.

I use the [Pillow](https://pillow.readthedocs.io/en/latest/index.html) library for Image support, and the way to write the import line is `from PIL import image`.  Pytest has no problem with that on my local Ubuntu system, but in Github, something's not right and it errors out with  Perhaps Pillow is not being loaded into the test environment?

ihabunek:
@danschwarz Thanks for the work. I didn't merge this for two reasons:
1. I didn't have the time to properly review this. I have things happening home which leave me with little spare time.
2. I'm always cautious when adding new dependencies to the project, and this added two.

I think we can drop ansicolor relatively easily. [24 bit color ](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit) is not hard to implement, similar to what is already done in [output.py](https://github.com/ihabunek/toot/blob/08dd02d9890f4ca240987e7944ab61463f213024/toot/output.py#L14-L26).

One request: please make a separate PR for scrolling? That's independant of images, and I think that can be merged right away. And then make a separate PR for images after. Thanks.

danschwarz:
I can do that. Scrolling can be a separate feature. Q: should both panes
have scrollbars or just status details side?

I have performance improvements that remove ansicolor so that’s not a
problem. Pillow is unavoidable.

PR #281, opened 2023-01-01
Make the status detail key commands more visible

danschwarz:
Some terminal color schemes completely eliminate the difference between cyan and cyan-bold colors (all the base16 themes, for instance). This change makes the key letters stand out clearly in bold white.


danschwarz:
On my system (Windows Terminal, Base16 theme)

Before:
image

After:
image

PR #282, opened 2023-01-03
Basic support for followed accounts in TUI

danschwarz:
In the status detail window, followed accounts are shown in yellow, while unfollowed accounts are shown in grey.
Not sure I love this UI representation (too subtle?) but it can be modified if desired.

Future possibility building on this: an [A]ccount option which will pop up an overlay with choices: Follow/Unfollow/Mute/Block.


ihabunek:
Thanks, merged manually.

PR #285, opened 2023-01-06
Support --help as the only command-line argument

ndw:
If you type  `toot --help`, you get the message 

```
Unknown command '--help'
Run toot --help to show a list of available commands.
```

which is a little, uh, disconcerting. This PR fixes that.


ihabunek:
Thanks!

PR #286, opened 2023-01-06
Make sure media are uploaded successfully

ndw:
The first time I tried to use toot, I attached an image to a message. The toot appeared, but without the image. Anecdotal evidence gathered from chatting with others about problems with media suggests that, especially when a server is busy, there can be a delay between when an upload is accepted and when it's actually available.

This PR adds a test-and-wait loop after uploading media to make sure they're available before continuing. It uploads the image and, if the response provides URIs, it loops for a couple of minutes waiting for them to be available. After a couple of minutes, it gives up and retries the upload. If the second attempt fails, it throws an exception.

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -1,5 +1,7 @@
 import re
 import uuid
+import time
+import requests
 
 from urllib.parse import urlparse, urlencode, quote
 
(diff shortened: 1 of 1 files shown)


ndw:
In subsequent testing, I discovered that this is a necessary but not sufficient test. It is possible for the API to accept the media, return media ids, for you to add those media ids to a toot, and yet for the published toot not to include the media. Possibly a race condition in Mastodon?

I ended up having to check that the status returned from the post *actually* contained all of the media ids it was supposed to. If it doesn't, I delete the toot and re-toot. But, alas, I didn't do that as a useful patch on toot.

ihabunek:
Thanks for your work. This is a Mastodon problem, but I agree we need to work around it.

I hesitate to merge this, as it slows down the happy path which in my experience works most of the time.

Could you test something please. When the status is posted the API returns the status object. When this issue occurs, does the returned status contain `media_attachments`? You can test this by adding `print(response)` after this command: https://github.com/ihabunek/toot/blob/b8f49ef2126b59f0418bde1ea57ba2a2d9d50a05/toot/commands.py#L95-L105

If it does not contain `media_attachments` for the posted media items, we could try using the new [edit](https://docs.joinmastodon.org/methods/statuses/#edit) feature to re-attach the media attachments. This way we would not affect the happy path, and only retry on failure.

ndw:
I'll see what I can do. It's definitely the case that after a successful media upload, you can get back a status response that doesn't include the media ids. I don't know if editing the response to insert them would work or not. I have *also* discovered that you cannot reuse media IDs. That is, you can't use the same media ID in different toots to refer to the same media. I don't know if an unsuccessfully attached media counts as "used" or not.

And, of course, the whole thing is difficult to test. I ended up spinning up my own Mastodon server in a docker container so that I could  post hundreds of stupid image toots in an effor to observe failure.

ihabunek:
Thanks. I also have a local mastodon server to run [integration tests](https://github.com/ihabunek/toot/blob/master/tests/test_integration.py).

ihabunek:
Just noticed that a v2 endpoint for uploading media exists https://docs.joinmastodon.org/methods/media/#v2

I wonder if using this instead of v1 would help.

ihabunek:
This has been fixed in 244502ec0b5c8dbb9ebfd5dc8174b858097e7b61 and released in v0.36.0. Please upgrade and test. And thanks for the contribution!

PR #288, opened 2023-01-09
Status detail scrollbar feature

danschwarz:
Uses scroll.py from https://github.com/rndusr/stig

Separated out from ANSI image preview feature.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-exclude=build,tests,tmp
+exclude=build,tests,tmp,toot/tui/scroll.py
 ignore=E128
 max-line-length=120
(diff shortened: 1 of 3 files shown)


danschwarz:
I have two ideas for what to do about the pane focus.

1. When the right pane has focus, change the color of the timeline status list to grey.  This could be done, I think, by wrapping the status_list in an AttrMap. 
2. Replace the vertical bar divider with \u258c Unicode Left Half Block. ( ▌)  Next, by controlling the foreground and background color of the divider characters, you can show two different colors in the single divider cell. Make the active side blue and the inactive side grey.   

The challenge with both of the above options is knowing when focus has shifted to the status detail pane and back again.  There's no on-focus or focus-lost signal that to use.  I think it could be added, but not sure of the best way to do it.  I tried leveraging SelectableColumns to do this but it's not built out to support something like Timeline widget yet.

PR #292, opened 2023-01-18
fix error during TUI startup

kaja47:
Version check failed when the server sent something other than a number as a version as happened on development version of the gotosocial server.


ihabunek:
Thanks!

Issue #188, opened 2021-05-28
Making a toot in TUI sends you back to the beginning of the timeline. Even if you reply to a thread.


newissuemaker:
This ruins the browsing experience and I can't daily drive toot because of this. It's good for posting a thing or two and checking notifications from cli but it's not usable for longer browsing sessions in TUI.

marathone:
Indeed, puzzling behaviour and the reason I don't use this app. I prefer to use a cli app but like @newissuemaker said, it ruins the experience!

danschwarz:
I think this is happening because the TUI rebuilds the timeline from scratch. It may be possible to save/restore the timeline position across timeline rebuilds.

PR #294, opened 2023-01-19
Fix for issue # 188

danschwarz:
ref: Issue #188


danschwarz:
Very strange... most of these tests succeeded, some failed and the failures appear to be GitHub infrastructure glitches, nothing related to the code itself. Would you rerun the tests? I think they will pass on second try.

ihabunek:
Thanks. You're right, restarting worked.

PR #296, opened 2023-01-20
Switch to showing relative times rather than full timestamps in status list

danschwarz:
You now see the seconds, minutes, days, or weeks since the toot was created, rather than the full timestamp.  A star is added to the time of toots that have been edited.

_The full timestamp is now shown in the status detail pane._

This change frees up 6 columns. These columns can be used to implement follow-on UI changes:
Scrollbar in status list, and focus boxes around the status list and status detail. (Box follows focus.)

image

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,4 @@ sphinx
 sphinx-autobuild
 twine
 wheel
+debugpy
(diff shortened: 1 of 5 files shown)


danschwarz:
Whoops - left my debugging code in place. I'll resubmit with that out.

PR #297, opened 2023-01-20
Status list now shows relative time created, rather than timestamp

danschwarz:
This is similar to how the stock Mastodon web client displays toot created_at times.  It's easier for users to interpret. A star is added for toots that have been edited (again, like the Mastodon web client.)   _The full timestamp is now displayed in the status detail pane_

This allows us to show the created_time delta in 4 columns rather than 10.  We can use those columns to add more UI features like a scrollbar for the status list, and a focus box that moves between the status list and status detail pane.

Here is what the UI looks like with the change:
image

Here is a preview of a future UI with status list scrollbar and focus box (NOT part of this pull request!).  Focus box is shown in both panes at once because I haven't figured out how to trigger on pane focus change (suggestions welcome!)

image

### DIFF
--- a/toot/tui/timeline.py
+++ b/toot/tui/timeline.py
@@ -10,6 +10,7 @@ from .utils import highlight_hashtags, parse_datetime, highlight_keys
 from .widgets import SelectableText, SelectableColumns
 from toot.utils import format_content
 from toot.utils.language import language_name
+from toot.tui.utils import time_ago
 
 logger = logging.getLogger("toot")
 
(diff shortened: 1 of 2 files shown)


ihabunek:
Thanks! Merged manually. I added a `--relative-datetimes` argument to enable this, since I personally prefer to have full datetimes.

Please check out the new [contributing guide](https://github.com/ihabunek/toot/blob/master/CONTRIBUTING.md), and especially the linked [article on commit messages](https://cbea.ms/git-commit/). No need to change existing pulls, but for the future, I'd appreciate if you could follow it. Thanks.

PR #298, opened 2023-01-21
Implement vim-like commands in footer

ihabunek:
Proof of concept, supports only `:quit` and `:help`.

### DIFF
--- a/toot/tui/app.py
+++ b/toot/tui/app.py
@@ -44,9 +44,13 @@ class Header(urwid.WidgetWrap):
 
 
 class Footer(urwid.Pile):
-    def __init__(self):
+    def __init__(self, tui):
+        self.tui = tui
         self.status = urwid.Text("")
         self.message = urwid.Text("")
+        self.command = Command(tui)
+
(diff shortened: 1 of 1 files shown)


ihabunek:
@danschwarz Playing around with vim style commands. This PR is a proof of concept, only implements `:quit` and `:help`. If you have some time check it out and see if it works well.

danschwarz:
Works fine; consider adding 

`self.loop.set_alarm_in(3, lambda *args: self.footer.clear_message())`

after

 `self.close(False, f"Unknown command: {command}")`

danschwarz (approved):
Works fine and should be expanded. 

Right now there are a lot of key bindings, some are non-intuitive. Could move the less frequently used commands to the vim interface and get them out of the key bindings. 

Consider adding

self.tui.loop.set_alarm_in(3, lambda *args: self.tui.footer.clear_message())

after

self.close(False, f"Unknown command: {command}")

ihabunek:
I decided not to do this in the end.

PR #299, opened 2023-01-22
Inline ANSI graphic image previews

danschwarz:
Submitting again after optimizations and removal of ansicolor dependency.
Supports truecolor and 256 color terminals (i.e. Apple Terminal). 
256 color requires use of a new --256 command line option.


danschwarz:
Tests failing due to lack of PIL import... which is the pillow library... which is included in requirements.txt.  Not sure how to fix.  Tests run fine locally with pillow installed.

ihabunek:
I wanted to merge this today, but didn't have the time. There are some changes I'd like to add before including this in master. Just letting you know I'm working on this, it looks really promising.

ihabunek:
Also on my TODO is refactoring the TUI to use asyncio, that would simplify everything...

I have a proof of concept [here](https://github.com/ihabunek/toot/blob/asyncio/toot/tui/async_app.py) but it requires a rewrite of the API layer, and replacing requests with aiohttp or another async lib...

danschwarz:
There’s one thing I’m slightly concerned with in the existing
implementation and that is the occasional slowdown in image loading after
you’ve browsed through 40 statuses or so. I think it’s due to garbage
collection. The statuses consume significantly more memory when you add the
image data to them. I wonder if it makes more sense to keep a single image
cache unconnected to the status objects with some kind of size limit and an
eviction policy? Single image cache would definitely speed up loading
avatar images, which frequently repeat across multiple statuses in a home
timeline.

danschwarz:
caught up to master

danschwarz:
Closing this out because I hope to supercede it with a version that will incorporate the term-image widget, providing the same functionality plus pixel-accurate images when using Kitty, WezTerm, or iTerm2 terminals. [(WIP [](here)](https://github.com/danschwarz/toot/tree/image11))

PR #300, opened 2023-01-26
Sixel display of avatars, media attachments, etc.

danschwarz:
I don't expect you to merge this PR into master, but it's fun to try out with a properly configured terminal that supports SIXEL images (Xterm works). 

~~You'll need to set fonts up as I do in .Xresources or things won't look correct.  I don't have autodetect of font-size-in-pixels yet, so it's hardcoded for these font settings.~~

Don't run under tmux or screen, unless you have a special build of those programs configured to pass through SIXELs.

image

.Xresources snippet 

`
! Use a truetype font and size.
! Cascadia Mono (no PL) should work as well
xterm*faceName: Cascadia Mono PL
xterm*faceSize: 14
xterm*background: black
xterm*foreground: grey

! Sixels support with a matching palette
XTerm*decTerminalID: vt340
XTerm*numColorRegisters: 256
`


ihabunek:
Nice, I need to find some time to try it out. I wouldn't necessarily be against merging it if it's guarded behind a user setting which is disabled by default.

danschwarz:
Yes, that makes sense. The code needs cleanup before a merge, particularly the bits in timeline.py need refactoring.

r3k2:
what this merged to main?

ihabunek:
Closing this one since toot tui is deprecated in favour of [tooi](https://codeberg.org/ihabunek/tooi) which supports sixel images.

PR #302, opened 2023-01-31
Added focus indicator box feature to the TUI

danschwarz:
Box follows focus: status list or status detail pane, left/right arrows or mouse clicks to change focus.  

Not sure if the commit message is any better or worse than before; it does explain "what" rather than "how" (although it doesn't say anything about "why")

image


danschwarz:
Closed for minor cleanup

PR #305, opened 2023-02-02
Add a save option to source dialog, tweaked by me

ihabunek:
This is a tweaked version of #304 

@danschwarz I made a bunch of changes, but the most significant is that instead of putting the button and filename edit box in the footer of a Frame, I put them in the ListBox alongside the JSON source. This means you can just scroll down to them and select them using arrow keys or pagedown.

Check out the changes and let me know what you think. I kept the commits small to make the changes simpler to review.


danschwarz:
The changes look good, I've made some additional changes on top of your changes - moving the filename and save button to the top of the list, so that they're immediately visible and available for use.  

I can't change this PR so I'm attaching a patch with my changes

image

[save.patch](https://github.com/ihabunek/toot/files/10574200/save.patch)

PR #306, opened 2023-02-03
Account feature - show account of current status author (/reblogger)

danschwarz:
Right now it's roughly at feature parity with `toot whois`

It can be extended to add commands (follow/unfollow, mute/unmute, etc.) if that's the right place to put those commands.  It is not tied strongly to a status item, so can be invoked from elsewhere, all that's needed as input is an account ID.

image

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -357,6 +357,10 @@ def followed_tags(app, user):
     return _get_response_list(app, user, path)
 
 
+def whois(app, user, account):
+    return http.get(app, user, f'/api/v1/accounts/{account}').json()
+
+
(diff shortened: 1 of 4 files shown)


ihabunek:
image

What witchcraft is this? :D I had no idea this is a thing, cool.

ihabunek:
Thanks!

ihabunek:
BTW, I changed this a bit to show the account of the poster of the toot, not the person who retooted it. Seems to make more sense.

danschwarz:
Yes, I saw that. Makes sense.

PR #308, opened 2023-02-04
Changed bookmark symbol from unicode glyph to 'b'

danschwarz:
The glyph previously used '🠷' is interpreted with different widths on different terminal emulators (i.e. xterm gets it wrong) This results in visual glitches in the status detail area and especially in the scrollbar when visible.  Changing it to a standard ASCII 'b' fixes all these issues.

Example of the glitch.  You can also see in this example that xterm doesn't render the 🠷 character at all, this is a shortcoming of the emulator itself (doesn't do fallback font substitution). 

image


ihabunek:
Thanks

PR #312, opened 2023-02-11
UI to vote in polls

danschwarz:
Basic UI to vote in polls

Changed a few key mappings to enable [P]oll key:

P -> E for Save (Pin) Timeline
E -> X for View Exception

Feel free to revert the key mapping changes and add your preferred mapping for Poll (Vote) functionality.

image

Needs a bit more testing for multi-selection polls, but single-selection polls are working fine.

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -1,4 +1,5 @@
 import re
+from typing import List
 import uuid
 
 from urllib.parse import urlparse, urlencode, quote
@@ -361,7 +362,7 @@ def whois(app, user, account):
     return http.get(app, user, f'/api/v1/accounts/{account}').json()
 
 
-def vote(app, user, poll_id, choices: list[int]):
+def vote(app, user, poll_id, choices: List[int]):
     url = f"/api/v1/polls/{poll_id}/votes"
     json = {'choices': choices}
     return http.post(app, user, url, json=json).json()
--- a/toot/api.py
+++ b/toot/api.py
@@ -361,7 +361,7 @@ def whois(app, user, account):
     return http.get(app, user, f'/api/v1/accounts/{account}').json()
 
 
-def vote(app, user, poll_id, choices: list):
+def vote(app, user, poll_id, choices: list[int]):
     url = f"/api/v1/polls/{poll_id}/votes"
     json = {'choices': choices}
     return http.post(app, user, url, json=json).json()
--- a/toot/api.py
+++ b/toot/api.py
@@ -361,6 +361,12 @@ def whois(app, user, account):
     return http.get(app, user, f'/api/v1/accounts/{account}').json()
 
 
+def vote(app, user, poll_id, choices: list):
+    url = f"/api/v1/polls/{poll_id}/votes"
+    json = {'choices': choices}
+    return http.post(app, user, url, json=json).json()
+
+
 def mute(app, user, account):
     return _account_action(app, user, account, 'mute')
 
--- a/toot/tui/poll.py
+++ b/toot/tui/poll.py
@@ -1,0 +1,98 @@
+import urwid
+
+from toot import api
+from toot.exceptions import ApiError
+from toot.utils import format_content
+from .utils import highlight_hashtags, parse_datetime
+from .widgets import Button
+
+
+class Poll(urwid.ListBox):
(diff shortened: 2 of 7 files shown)


toot/api.py
```diff
@@ -361,6 +361,12 @@ def whois(app, user, account):
     return http.get(app, user, f'/api/v1/accounts/{account}').json()
 
 
+def vote(app, user, poll_id, choices: list):
```
ihabunek, line 364: Since you're adding types, specify what should be in the list. Looking at the API, it's integers so `List[int]`.

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def build_linebox(self, contents):
+        contents = urwid.Pile(list(contents))
+        contents = urwid.Padding(contents, left=1, right=1)
+        return urwid.LineBox(contents)
+
+    def vote(self, button):
+        poll = self.status.data.get("poll")
+        choices = []
+        for idx, b in enumerate(self.button_group):
```
ihabunek, line 35: I prefer having more verbose variable names, unless it's a problem with line width. So `button` instead of `b` here makes it a bit more readable.

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+            if b.get_state():
+                choices.append(idx)
+
+        if len(choices):
+            try:
+                response = api.vote(self.app, self.user, poll["id"], choices=choices)
+                self.status.data["poll"] = response
+                self.poll["voted"] = True
+                self.poll["own_votes"] = choices
```
ihabunek, line 44: Would this work as well?

```suggestion
                self.poll = response
```

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+                    "%Y-%m-%d %H:%M"
+                )
+                poll_detail += " · Closes on {}".format(expires_at)
+
+            yield urwid.Text(("gray", poll_detail))
+
+    def generate_contents(self, status):
+        yield urwid.Divider()
+        for line in format_content(status.data["content"]):
+            yield (urwid.Text(highlight_hashtags(line, set())))
```
ihabunek, line 89: Unnecessary parens: 
 
```suggestion
            yield urwid.Text(highlight_hashtags(line, set()))
```

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+
+        yield urwid.Divider()
+        yield (self.build_linebox(self.generate_poll_detail(status)))
```
ihabunek, line 92: Same.
```suggestion
        yield self.build_linebox(self.generate_poll_detail(status))
```

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+                self.poll["own_votes"] = choices
+            except ApiError:
+                pass  # FIXME error reporting?
+            finally:
+                self.setup_listbox()
+
+    def generate_poll_detail(self, status):
+        poll = self.poll
+
+        if poll:
```
ihabunek, line 53: Do we need this check? This widget should never get constructed if there is no poll.

If we do, I'd prefer something like:

```python
if not poll:
    return
```

so that we don't have an extra indent for the rest of this method.

toot/tui/poll.py
```diff
@@ -0,0 +1,98 @@
...
+                choices.append(idx)
+
+        if len(choices):
+            try:
+                response = api.vote(self.app, self.user, poll["id"], choices=choices)
+                self.status.data["poll"] = response
+                self.poll["voted"] = True
+                self.poll["own_votes"] = choices
+            except ApiError:
+                pass  # FIXME error reporting?
```
ihabunek, line 46: Maybe add a text box to display errors to the bottom of the listbox?

ihabunek (changes requested):
Thanks. This mostly looks good, I have some nitpicks in the comments.

It currently does not allow you to select multiple options if `poll.multiple = True`. This would be nice to add before merging.

ihabunek:
Let me know when this is ready to review by re-requesting the review in the top right.

ihabunek:
I merged this since the code looks good.

PR #313, opened 2023-02-17
Implement classes for API entities

ihabunek:


danschwarz:
This looks like it's going to be good. One thing to consider; managing reblogged statuses are a source of potential bugs (referencing the reblog when you mean to be referencing the original status, or vice-versa).  For instance, I think there's a bug with the [L]ink command finding links in reblogged statuses right now, but haven't put in the time to find out what's going on there yet.

Anything that can be done to make it very easy to reference the proper status (reblog vs original) may help. I think that most use cases reference the original status, not the reblog, so defaulting to the original status may be best.

ihabunek:
It definitely is a common source of bugs. That said, not sure what I can do in addition to having the [`original` property](https://github.com/ihabunek/toot/blob/414a7fc47a4e8674a542330ecf5236e347455501/toot/tui/entities.py#L249-L250) which returns the reblog if it's reblogged, or the base status otherwise - i.e. it returns the status which contains the contents we want to display. Perhaps it's not the best name?

You need to be able to access both statuses, since you need them for various use cases. E.g. when displaying a reblogged status you need the main status to see who reblogged it.

danschwarz:
Seems to me that the only case you want the reblog status info is when
displaying the reblog header. I wonder if it makes sense to flip how it’s
accessed:  “unwrap” the original status so it is accessed directly in all
cases, and the reblog status data is only accessed from a status.reblog
property ?

ihabunek:
Well the classes in `entities.py` need to be 1:1 with API entities if I want to have this automatic creation from the data returned by the API. But we can definitely have a custom `Status` class for the TUI to use, or something similar. I'll work on finishing this first, and then we can improve.

ihabunek:
Replaced by #374

PR #315, opened 2023-02-21
Account update

### LINKED ISSUE #314: Feature Request: Profile management
hello toot friends 

Please in some way find time to add the ability to update the profile picture, profile banner and profile text via toot command line arguments.

ihabunek:
Implements updating account details.

I'm tempted to rename the command from `toot update_account` to `toot account` to keep it short, but it may be a bit ambiguous. :thinking: 

fixes #314


danschwarz (approved):
Changes look good to me. I would not change the name, because these are likely to be infrequently used commands.

PR #317, opened 2023-02-22
Account overlay - don't allow mute/block/unfollow own account

danschwarz:
Realized that it doesn't make sense to offer the user the option to follow, mute, or block their own account  🙄 

image


ihabunek:
Thanks. BTW you can clone the toot repo and push branches directly to it if you want. Makes it so you don't need to maintain a fork.

danschwarz:
Thanks. I think I’ll keep my fork for now because I have code in progress
that I want to keep in the cloud, but I don’t want to clutter the main
repo.

PR #320, opened 2023-02-25
Basic support for viewing Status and Mention notifs as a timeline

danschwarz:
Now the Goto Menu has a < Notifications > button. This will load your Status and Mention notifications as a standard timeline for viewing.

image

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -290,8 +290,11 @@ def _timeline_generator(app, user, path, params=None):
 def _notif_timeline_generator(app, user, path, params=None):
     while path:
         response = http.get(app, user, path, params)
-        notif = response.json()
-        statuses = [n['status'] for n in notif]
+        notification = response.json()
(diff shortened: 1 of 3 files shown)


toot/api.py
```diff
@@ -290,8 +290,11 @@ def _timeline_generator(app, user, path, params=None):
 def _notif_timeline_generator(app, user, path, params=None):
     while path:
         response = http.get(app, user, path, params)
-        notif = response.json()
-        statuses = [n['status'] for n in notif]
+        notification = response.json()
+        statuses = []
+        for n in notification:
+            if n['status']:
+                statuses.append(n['status'])
```
ihabunek, line 297: Pro tip, you can add the if clause inside the list comprehension:

```py
statuses = [n['status'] for n in notif if n['status']]
```

ihabunek:
Thanks!

PR #321, opened 2023-02-28
Browse other servers' public local timelines

danschwarz:
Here's an interesting feature ~~that's about 3/4 done (so this is another draft).~~

With this PR you can select another Mastodon server and browse its local public timeline, anonymously.  For example, if you're interested in the Rust language, [hachyderm.io](https://hachyderm.io) has a lot of relevant content.

image

While you browse, you're still logged in to your active account on your home server.

image

The previously reported bugs have been squashed, per my limited testing.

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -266,6 +266,11 @@ def context(app, user, status_id):
     return http.get(app, user, url).json()
 
 
+def anon_context(instance, status_id):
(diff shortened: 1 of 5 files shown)


danschwarz:
It's actually working pretty well now. Code suggestions, bug reports appreciated when you have time.

danschwarz:
Closed in favor of #323

PR #323, opened 2023-03-01
Better, more complete support for browsing public local timelines on "foreign" servers

danschwarz:
This seems to be working well now. 

Appreciate any testing/review when you have the time.

image

image


danschwarz:
Note that Glitch fork Mastodon servers do not allow anonymous browsing of public timelines (this may be a configurable feature on their server, but the two I've tested, including glitch.social, throw a 422 error (validation).

danschwarz:
Bug fix on the way for this PR - crash when browsing a foreign server with LOCAL_DOMAIN set.

danschwarz:
> Bug fix on the way for this PR - crash when browsing a foreign server with LOCAL_DOMAIN set.

done.

danschwarz:
Closing this PR in order to roll back any commits related to support of servers that have custom LOCAL_DOMAINS. When PR #335 lands in master, I'll merge that with the public local timeline support and see about resubmitting a new PR.  Browsing other servers' public local timelines is one of the features I use most frequently now, and it's a feature that a lot of other Mastodon clients lack.

danschwarz:
Well, merging with master was no problem, I've rolled back all the local domain code that is no longer needed, so we're back to the original submission.

There might be a nicer/cleaner way to implement this; I'm open to suggestions.  But I recommend trying it out, if only for the cleaned up Goto Menu, which is quite nice.

image

danschwarz:
I can't reopen this.

PR #319, opened 2023-02-25
Latest version of image support for toot

danschwarz:
Currently supports Unicode half-cell rendering, plus Kitty and iTerm2 pixel rendering in supported terminals. Sixel support is a future planned enhancement. (Thanks @AnonymouX47 !) 

I've done some cleanup and simplification of the code. 

image

image

Issue #324, opened 2023-03-02
Sixels or Unicode image previews


osresearch:
![timg](https://user-images.githubusercontent.com/3068843/222464582-246645f6-5351-4cd4-85d2-f3d32c535da0.jpg)

For toots with media attachments, it might be interesting to use [sixels](https://en.wikipedia.org/wiki/Sixel) or [timg](https://github.com/hzeller/timg) to render them in the reader pane with unicode blocks.

osresearch:
![image](https://user-images.githubusercontent.com/3068843/222482674-e7f0e9cf-e7ad-4576-bf0e-3ea7d97398b6.png)

`urwid` is eating the foreground/background attributes, but the `timg` based decoder is almost working

osresearch:
![image](https://user-images.githubusercontent.com/3068843/222487262-03ba6c69-9045-4e2e-9d39-f3c77a6274b6.png)

Oops, it was a problem with the distro's ancient version of urwid.  Installing via `pip` and now it "works".

ihabunek:
There is ongoing work to implement this, see:
https://github.com/ihabunek/toot/issues/243
https://github.com/ihabunek/toot/pull/319
https://github.com/danschwarz/toot/discussions/2

danschwarz:
The images branch is now working quite well IMHO and cleanly separates the image related code into a single file; you can choose to install toot with or without image support (image support requires the term-image and pillow libraries),

pagiraud:
I know there has been a lot of progress on this side, and images are rendered quite well with Kitty, for instance. But, to be honest, I don’t see the point of the block mode, but if it is to have an idea of what some people with eyes impairment can feel.
Is it planned to implement a sixel mode, so that images can be rendered on Foot terminal, for instance?

PR #325, opened 2023-03-02
Render images attachments into unicode blocks

osresearch:
image

This patch adds very hacky support for invoking `timg` to render image attachemnts into the terminal.  There is a probably a better way to do this, such as https://github.com/ihabunek/toot/pull/319

### DIFF
--- a/toot/tui/timeline.py
+++ b/toot/tui/timeline.py
@@ -15,6 +15,54 @@ from toot.tui.utils import time_ago
 
 logger = logging.getLogger("toot")
 
+import subprocess
+import re
+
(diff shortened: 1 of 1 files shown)


ihabunek:
Thanks for your contribution! This is a nice proof-of-concept, but I'm rejecting it because there is a much more comprehensive effort in progress, as mentioned in https://github.com/ihabunek/toot/issues/324#issuecomment-1452149986.

osresearch:
Thanks for `toot`! Yeah, this was not intended to be merged; more of a "look, it works!" proof-of-concept.

Issue #301, opened 2023-01-27
Feature request: image preview on "toot post 'x' --media


carlmalamud:
Toot post allows one to attach a media file and a description. But, there is no provision to upload a preview image for the attachment, which is very handy if you are uploading audio or video. Any chance we could get --preview [image.jpg] as an additional option?

Thanks for toot ... very useful! 

Carl

danschwarz:
fixed as of 0.36 if not earlier

PR #326, opened 2023-03-03
Add media thumbnails

ihabunek:
This turned out to be a bit more complex than expected. Had to implement logic to wait for all media to finish processing.

issue #301


toot/console.py
```diff
@@ -482,6 +482,12 @@ def editor(value):
                 "help": "plain-text description of the media for accessibility "
                         "purposes, one per attached media"
             }),
+            (["--thumbnail"], {
```
danschwarz, line 485: Should this --thumbnail flag be added to the `toot upload` command as well?

danschwarz, line 485: Should this `--thumbnail` argument be added to the `toot upload` command as well?

toot/commands.py
```diff
@@ -142,19 +143,54 @@ def _get_scheduled_at(scheduled_at, scheduled_in):
...
+    has finished processing it. This function waits for that to happen.
+
+    Once media is processed, it will have the URL populated.
+    """
+    if all(m["url"] for m in uploaded_media):
+        return
+
+    # Timeout after waiting 1 minute
+    start_time = time()
+    timeout = 60
```
danschwarz, line 175: This is a troublesome Mastodon API; you are forced to poll indefinitely and have no idea how long the server will need to process the media. 60 seconds is reasonable, but to account for large files or slow servers, should this be (yet another) config option?

danschwarz (approved):
Looks good- tested it with a small video file and thumbnail, worked as expected with all options. See my comments but they should not block merging.

toot/commands.py
ihabunek: Yeah, it would be good to be able to customize this. Once I finally add a settings file... Although i think 1 minute should be plenty in the vast majority of cases.

toot/console.py
ihabunek: Yeah, it probably should. I'll take care of it later.

PR #327, opened 2023-03-03
Display "New notifications" indicator in TUI header when a user has new notifications

danschwarz:
Displays a "New Notifications" indicator when new notifications
are found. This indicator is cleared when the user opens the
notifications timeline.

The code is currently configured to check for new notifications
every 2 minutes.

The notification id of the last notification viewed is stored
in the config file; example:

image

image

Tip: for easy testing of the "new notifications" indicator, reset your toot config.json file with this one-liner: 

`sed -i 's/\"last_id\".*/\"last_id\": \"0\"/g' ~/.config/toot/config.json`


toot/api.py
```diff
@@ -289,7 +289,19 @@ def _notification_timeline_generator(app, user, path, params=None):
...
+        statuses = []
+        for n in notification:
+            status = n["status"]
+            # hack to have notification info available in our statuses
+            # so we don't have to redo all interfaces.
+            # preface with toot_ so we don't collide
+            # with real status tags now or in future
+            status.setdefault('toot_notification_id', n.get('id'))
+            status.setdefault('toot_notification_type', n.get('type'))
+            status.setdefault('toot_notification_created_at', n.get('created_at'))
```
ihabunek, line 302: Hm, I dislike this hack, since it's polluting the `api` module with things that only concern the TUI. I think we should hold off on this until I get a chance to finish #313, then we can put this in `StatusMeta`.

danschwarz:
Ok

PR #330, opened 2023-03-04
Add conversations support in TUI

rogarb:
Using the model of notifications, I added support to conversations in the TUI.


ihabunek (approved):
Good work, thanks.

Issue #331, opened 2023-03-05
[TUI] Default visibility in reply should be the same as in replied toot


rogarb:
When replying to a toot using the TUI, the visibility in the reply is set to public by default. IMO the visibility should remain the same as the toot which is replied to, e.g. a reply to a "direct" toot should have a "direct" visibility by default.

PR #332, opened 2023-03-05
Inherit visibility when replying in TUI

rogarb:
When replying to a toot, the visibility gets set to default: if the toot has a direct visibility, the visibility in reply will be changed to the default one, whereas it should stay in direct mode.

This fixes the issue #331.


ihabunek:
Thanks!

PR #334, opened 2023-03-07
Add cop[y] status feature - copies status text to clipboard

danschwarz:
This relies on the OSC 52 terminal feature, which is widely supported (Windows Terminal, iTerm2, XTerm, Kitty, others). Basic support for copying the entire status text content is in place- but this can be improved in future by removing HTML tags, adding URL links, author info etc. as is done for status messages in the `toot timeline console command.


ihabunek (approved):
Nice, thanks.

PR #335, opened 2023-03-07
Add support for custom instance domains

ihabunek:
The instance domain can be different from their base url, for example the instance at https://social.vivaldi.net/ uses the vivaldi.net domain, sans 'social'.

This commit requires the user to provide the base url of the instance, instead of domain name. The domain is then fetched from the server.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -14,7 +14,7 @@ def test_create_app(mock_post):
         'client_secret': 'bar',
     })
 
-    create_app('bigfish.software')
+    create_app('https://bigfish.software')
 
     mock_post.assert_called_once_with('https://bigfish.software/api/v1/apps', json={
         'website': CLIENT_WEBSITE,
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -13,17 +13,18 @@ def test_register_app(monkeypatch):
         assert app.client_secret == "cs"
 
     monkeypatch.setattr(api, 'create_app', retval(app_data))
-    monkeypatch.setattr(api, 'get_instance', retval({"title": "foo", "version": "1"}))
+    monkeypatch.setattr(api, 'get_instance', retval({"title": "foo", "version": "1", "uri": "bezdomni.net"}))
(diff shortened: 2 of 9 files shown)


ihabunek:
> Note that the Mastodon api get instance has a v2 version available since 4.0.0 that returns a domain tag. This is the LOCAL DOMAIN value. More reliable than looking at base url.

v1 also contains the domain, it's just called `uri`. If you have a look [at the source](https://github.com/mastodon/mastodon/blob/main/app/serializers/rest/v1/instance_serializer.rb#L15-L17) you can see uri is populated with the domain.

I choose to keep v1 because v2 was introduced relatively recently, and v1 will probably remain supported for a good while.

danschwarz:
I agree with your approach on the get instance API.  

Unfortunately the code does not work at all to log in to my account on https://social.vivaldi.net .  If I use a login file created with the master branch, log in to social.vivaldi.net, then manually edit config.json, changing base_url from https://social.vivaldi.net to https://vivaldi.net,  I can login but we crash on get_instance. Haven't debugged past that point.

image

As far as I can tell, the rules with LOCAL_DOMAIN are:
All communication with the server needs to use the instance 
But the local account names should use LOCAL_DOMAIN rather than the instance
Toots should be displayed with LOCAL_DOMAIN, account lookup should use LOCAL_DOMAIN for local accounts.

My patch is messier but accounts for most of this.  It's not 100% yet.

ihabunek:
How does it not work? I did my testing and it seems to work fine. 
Here's my notes: https://gist.github.com/ihabunek/4c920f035cbefced926f7a17b5c35083

> then manually edit config.json, changing base_url from https://social.vivaldi.net/ to https://vivaldi.net/

You should not change `base_url`, the server is located at https://social.vivaldi.net/. We just want to change the `instance`.

danschwarz:
I’ll test later with your gist instructions, thanks

danschwarz (approved):
Tested again and it works well. One possible add-on to consider in future:  For any existing logins where we can detect a custom domain/uri, fix the values in config.json.

PR #336, opened 2023-03-08
Add polls to toot post

ihabunek:


danschwarz (approved):
Works well. Could form the basis of a future TUI feature to compose toots with polls.

Issue #337, opened 2023-03-09
Refresh ',' on any timeline other than Home, puts you back on the home TL


danschwarz:

danschwarz:
Closed as fixed by commit [999d97](https://github.com/ihabunek/toot/commit/9999d975b4b3b6d0bb2f6c8e3100157309c3916f)

PR #338, opened 2023-03-09
Fixed refresh so it stays on the currently selected timeline

danschwarz:
Fixes issue #337


danschwarz:
As a side effect of this change (I consider it a positive) the public timelines are now differentiated in the footer: [local public] and [global public]

ihabunek:
Thanks. Merged manually. 

For the future, I have an idea for objects which encapsulate the status generators and which have a `rewind` method. That would make this much simpler.

danschwarz:
That `rewind` method ... will it prepend new toots to the top of a timeline being refreshed? Or does `rewind` mean something else?

PR #342, opened 2023-03-15
Don't use signals

ihabunek:
This PR removes signals in favour of direct invocation. I don't _think_ we loose anything by doing this and it simplifies the code. WDYT?


danschwarz:
Agree that this will simplify things.  I'm not sure what benefit the signals provided.  This will also allow us to refactor app.py, moving functionality out that doesn't need to be there.

PR #343, opened 2023-03-17
Fix [b]oost so it works on statuses that were also boosted by others

danschwarz:
Boosting of statuses that others have already boosted, doesn't work properly in the TUI (try it and watch the yellow boost flag; it doesn't toggle on/off as you would expect).

These changes properly refer to the original status when boosting / unboosting, and when creating the new status for display. You can verify that toggling works correctly.  Boosting of non-boosted statuses also continues to work correctly.


danschwarz:
Something's weird with the code pushed in this PR; closing and reopening a new one with clean code

PR #348, opened 2023-03-24
Support to display a limited set of HTML tags (aligned with mastodon PR #23913)

danschwarz:
I started with GPL3 licensed code from https://github.com/seonon/muv, did a lot of cleanup and rewriting.

Handles all the tags specified in https://github.com/mastodon/mastodon/pull/23913

Combining EM and STRONG, or B and I tags in any combination now works.

### How rich text looks on the web client of a glitch-soc server:
image

### The same status on Toot
image


danschwarz:
This version of the code is the first one I'd consider "good enough to criticize".  Try it out and see what you think. You can find examples of Mastodon statuses with embedded HTML on Pleroma and Akkoma servers (or create one yourself using Markdown).

Feature-complete for rendering of all tags supported by Mastodon (current and upcoming PR#23913), Pleroma, Akkoma.

Now supports automatic highlighting of #hashtags (via the class on the hashtag anchor tag).  The subfeature of highlighting followed hashtags in yellow is not supported and I don't plan to add it.

Needs:

* Comparison vs current state on display of statuses - ideally, existing statuses should display identically.
* Opinions on how HTML tags should be rendered - whitespace above/below block tags, * versus unicode • for unordered <LI> items,  how to best display <PRE> and <CODE> blocks, how to render <H1>..<H6>.
* Thorough automated test suite against a wide variety of HTML statuses - with valid and invalid HTML, supported and unsupported tags.

danschwarz:
One design limitation of the code is that it maps a tag or an HTML class to an urwid palette constant name. So I have defined palette entries for most HTML tags.
```    
    ('a', ',italics', ''),
    ('em', 'white,italics', ''),
    ('i', 'white,italics', ''),
    ('strong', 'white,bold', ''),
    ('b', 'white,bold', ''),
    ('u', 'white,underline', ''),
    ('del', 'white, strikethrough', ''),
    ('code', 'white, standout', ''),
    ('pre', 'white, standout', ''),
    ('blockquote', 'light gray', ''),
    ('h1', 'white, bold', ''),
    ...
    ('class_mention_hashtag', 'light cyan,bold', '')
```
It has no way to map to _combinations_ of tags. So if you have <b> nested in an <i> tag, you should get bold and italic text, but in the current implementation, you will only get bold text. There may be a way to work around this using the current design, say, to define a 'bi' palette entry for both bold and italic, and find a way to detect and return a 'bi' attribute name when nested tags like this are detected. I haven't coded that yet.  
There are more elegant solutions that would require extensive redesign.

danschwarz:
Switched this PR back to draft because of a rendering issue that I can't fix. There may be an underlying rendering bug within `urwid.Text` causing the problem.

The issue is that, in certain cases, all text to the right of an anchor tag is not displayed.  Text on subsequent rows is displayed.  If you resize the terminal window such that the text to the right of the anchor tag wraps around to the next row, it does display properly on the next row.



I have a separate branch where I have reimplemented support for rich text using @AnonymouX47 's `TextEmbed` and `Hyperlink` widgets, and those widgets do render correctly.  The code is signficantly more complex due to issues around hashtag highlighting (requires interpretation of HTML class attributes within nested tags). Testing it to make sure it's robust enough before submitting as a new PR.

danschwarz:
oh, and `TextEmbed` and `Hyperlink` are in urwidgets, which is currently a development release only; we'll have to wait for a real release before those components can be used here.

AnonymouX47:
> The issue is that, in certain cases, all text to the right of an anchor tag is not displayed.

That's a really weird one but really I suspect the terminal emulator in this case.
I suppose this is Windows Terminal. Have you tried any other terminal emulator?

> we'll have to wait for a real release before those components can be used here.

Will be done this weekend.

danschwarz:
>really I suspect the terminal emulator in this case.
I suppose this is Windows Terminal. Have you tried any other terminal emulator?

Yes, unfortunately the same problem occurs when testing with kitty and xterm.js.
It's possible I'm constructing the markup for urwid.Text in some way that is invalid, but in that case I would expect a crash or a failure to display the text in all cases - not just when it appears on the same line as the link.

>I haven't seen your code but I remember I wanted to suggest using urwidgets.parse_text() recursively (i.e calling parse_text within repl). I had intentions to work out a sample for inline tags (bold, italics and underline) but I guess I got caught up in other things and forgot.

I'd be interested in seeing this sample if you have time to write it, although I am not sure it will be readily applicable to this use case.

>Will be done this weekend.

Terrific!

AnonymouX47:
> Yes, unfortunately the same problem occurs when testing with kitty and xterm.js.

Ouch! That doesn't sound good :\.

> It's possible I'm constructing the markup for urwid.Text in some way that is invalid

Possibly 🤔...

> I'd be interested in seeing this sample if you have time to write it, although I am not sure it will be readily applicable to this use case.

Interestingly, I just finished it up right now. Was about to post it here when I saw your reply :grin:.

All of it may not be directly applicable being that it only covers a subset of the HTML elements concerned but I think the same approach can be applied and even some of the code can be useful. I'll post it in my next comment.

AnonymouX47:
The main code:

```py
import re
from urwidgets import parse_text

INLINE_HTML_TAGS = {
    "b": ("b", "strong"),
    "i": ("i", "em"),
    "u": ("u",),
    "s": ("s", "del"),
}
INLINE_HTML_PATTERNS = {
    re.compile(f"<(?:{'|'.join(tags)})>(.*?)</(?:{'|'.join(tags)})>"): attr
    for attr, tags in INLINE_HTML_TAGS.items()
}

SAMPLE_HTML = """\
<em>emphasized</em>
<u>underlined</u>
<b>bold</b>
<b><i>bold and italic</i></b>
<del>strikethrough</del>
regular text
<u><b>underlined and bold <i><s>underlined, bold, italic and strikethrough</s></i></b> only underlined</u>
"""


def inlne_html_repl(pattern, groups, span, attr):
    attr = attr + INLINE_HTML_PATTERNS[pattern]
    markup = parse_text(groups[1], INLINE_HTML_PATTERNS, inlne_html_repl, attr)

    return (
        markup
        if (
            isinstance(markup, tuple)
            or isinstance(markup, list)
            and all(isinstance(element, tuple) for element in markup)
        )
        else ("".join(sorted(attr)), markup)
    )


markup = parse_text(SAMPLE_HTML, INLINE_HTML_PATTERNS, inlne_html_repl, "")
```

With
```py
if __name__ == "__main__":
    print(*INLINE_HTML_PATTERNS.items(), sep="\n")
    print()
    print(markup)
```

I get
```
(re.compile('<(?:b|strong)>(.*?)</(?:b|strong)>'), 'b')
(re.compile('<(?:i|em)>(.*?)</(?:i|em)>'), 'i')
(re.compile('<(?:u)>(.*?)</(?:u)>'), 'u')
(re.compile('<(?:s|del)>(.*?)</(?:s|del)>'), 's')

[('i', 'emphasized'), '\n', ('u', 'underlined'), '\n', ('b', 'bold'), '\n', ('bi', 'bold and italic'), '\n', ('s', 'strikethrough'), '\nregular text\n', ('u', [('bu', ['underlined and bold ', ('bisu', 'underlined, bold, italic and strikethrough')]), ' only underlined']), '\n']
```
as output.

With
```py
from itertools import combinations
import urwid

if __name__ == "__main__":
    def handle_input(key):
        if key == "q":
            raise urwid.ExitMainLoop

        return False

    attributes = ('bold', 'italics', 'strikethrough', 'underline')
    palette = [
        ("".join(attribute[0] for attribute in combo), ",".join(combo), "")
        for combo in [
            tuple(combo)
            for n in range(1, len(attributes) + 1)
            for combo in combinations(attributes, n)
        ]
    ]
    print()
    print(*palette, sep="\n")

    text = urwid.Text(markup)
    base_widget = urwid.Filler(text)

    loop = urwid.MainLoop(base_widget, palette=palette, unhandled_input=handle_input)
    loop.run()
```

I get
```
('b', 'bold', '')
('i', 'italics', '')
('s', 'strikethrough', '')
('u', 'underline', '')
('bi', 'bold,italics', '')
('bs', 'bold,strikethrough', '')
('bu', 'bold,underline', '')
('is', 'italics,strikethrough', '')
('iu', 'italics,underline', '')
('su', 'strikethrough,underline', '')
('bis', 'bold,italics,strikethrough', '')
('biu', 'bold,italics,underline', '')
('bsu', 'bold,strikethrough,underline', '')
('isu', 'italics,strikethrough,underline', '')
('bisu', 'bold,italics,strikethrough,underline', '')
```

and

Screenshot_2023-05-12_14-50-33

as output.

AnonymouX47:
> > we'll have to wait for a real release before those components can be used here.
> 
> Will be done this weekend.

https://pypi.org/project/urwidgets/

Kept to my word for once! 🥲😁

danschwarz:
and I have the PR incorporating it almost ready to go...

danschwarz:
Thanks for the sample code. Your sample supports even more font style combinations than Mastodon will allow in their 4.2 release.

danschwarz:
Closing this to open a new PR...

PR #356, opened 2023-05-15
Support to display a limited set of HTML tags (aligned with mastodon PR #23913) plus URL support

danschwarz:
This is essentially a rewrite of the closed PR #348 using @AnonymouX47 's `TextEmbed` widget from `urwidgets`. It does not suffer from the rendering issue that caused problems for the previous PR, that used `urwid.Text`.   As a bonus, it includes support for clickable links on terminals that support OCS 8 hyperlinks.

This PR renders all HTML elements, and some HTML attributes, that will be supported in Mastodon 4.2 [spec](https://docs.joinmastodon.org/spec/activitypub/#sanitization)

### HTML Status on a glitch-soc server
image

### The same status on the TUI with this PR
image

As the `urwidgets` dependency is Python 3.7+,  I have removed the Python 3.6 automated Github tests.


README.rst
danschwarz, line 1: No idea why this file is included in the PR. The version in the PR is identical to the version in master currently.

danschwarz:
Update- I've got a new version of this PR that removes the hard dependency on the urwidgets library.  If you have urwidgets installed, it works with full functionality including OCS 8 support for hyperlinks.  If you do not have urwidgets, it still works, all HTML tags are shown properly, but hyperlinks display as plain text.  When it's good enough to submit, I'll close out this PR and open up a new one.

danschwarz:
Closed in favor of #403 PR which is the same code as above, without the hard dependency

Issue #354, opened 2023-05-10
Feature Request: personal timeline


Rostiger:
An additional timeline that displays all your posts/threads to better keep track of conversations.

PR #357, opened 2023-05-18
Personal timeline in TUI and toot timeline --account_tl user@instance

danschwarz:
This addresses issue #354 .
Lightly tested. Works.
I'm not thrilled with the names, so feel free to reword as you like.

image

image


ihabunek:
Thanks! I merged this manually (had to rebase). Mostly good, but I changed the commandline options so that instead of having both `--account_tl` and `account` we have just one option: `--account ihabunek@mastodon.social` .

PR #358, opened 2023-05-26
Login to servers that don't honor the uri spec for V1::Instance

### LINKED ISSUE #347: toot login gets the instance info wrong when logging into Akkoma server
`toot login`

`Enter instance URL [https://mastodon.social]: https://blob.cat
Looking up instance info...
Found instance Blob.cat running Mastodon version 2.7.2 (compatible; Akkoma 3.3.1-0-gbba467b39-blob)`

... go through the login process. ...
... check config.json ...

Note the instance is `https://blob.cat`

image

This is with the latest master.
I have confirmed that doing the same process against a stock Mastodon server results in the correct instance data being stored in config.json (no https:// prefix)

Not sure whether this is the fault of Akkoma, or toot, yet.

danschwarz:
Pleroma, Akkoma, and other servers do not follow the Mastodon spec for the 'uri' attribute which specifies that it contains the domain name of the instance. Instead, they return a complete URI.

As a workaround, we now detect this situation and parse out the domain from the URI when necessary. Fixes #347 

Thanks to @laleanor for their patch and @rjp for ideas on how to make it work with GotoSocial and other servers

### DIFF
--- a/toot/auth.py
+++ b/toot/auth.py
@@ -7,6 +7,7 @@ from getpass import getpass
 from toot import api, config, DEFAULT_INSTANCE, User, App
 from toot.exceptions import ApiError, ConsoleError
 from toot.output import print_out
+from urllib.parse import urlparse
 
 
 def register_app(domain, base_url):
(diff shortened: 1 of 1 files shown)


ihabunek:
This looks sensible. Thanks! 

@danschwarz Sorry for my absence. I'm fine, just got hit by work and personal life issues at the same time. I saw your IRC messages but you were offline so couldn't reply. Feel free to hit me up with email or Mastodon DM if you want to talk.

danschwarz:
Hey, no worries, welcome back, I figured it was something like that. I may need to step away for a while to get some unrelated stuff done, too.  Meanwhile, I hope the PRs with dozens of commits are not a nightmare to review.

PR #359, opened 2023-05-26
Rich text support in console output

danschwarz:
This PR is cumulative with PR #356 . It adds the ability to output rich text as Markdown on the console (i.e., in `toot timeline` output) and also output rich text as Markdown to the clipboard with the `Cop[y]` command.

It uses the [html2text ](https://pypi.org/project/html2text/)library to convert HTML to Markdown text.

Why Markdown format? Because there doesn't seem to be a html2ansi library available anywhere.  Also, ANSI escape sequences aren't a clipboard-friendly format if we want to copy richtext status messages to the clipboard using OSC 52.

Why this library? Because I couldn't find anything better in a prebuilt library. I'm not interested in writing all the code to do text formatting, wrapping, etc. when all that has been done here.

Output of ordinary status messages without HTML passes the console_test.py test suite.  I have added an additional test of a status message with all the new HTML tags to be supported in Mastodon 4.2.

That said, html2text is not great! It has some formatting bugs and open PRs piling up since 2022.  The last update in Github was February 2022. You can see in the example below, lines with emphasized, bold, and combined styles have extra spaces in them.  I have a [fork of the project ](https://github.com/danschwarz/html2text)where I've applied all the relevant PRs and it fixes *some* of the formatting bugs, but not *all*.  (The output you see below uses the release version of html2text, not my fork).

Open to suggestions about how to proceed with this. 

image


danschwarz:
> That said, html2text is not great! It has some formatting bugs and open PRs piling up since 2022.

I've fixed the one HTML2Text issue (design choice) that stopped long URLs from being properly wrapped at the proper width. This is only in my fork of the project for the foreseeable future as upstream is dormant.

The issue with spaces being added in front of underlined text appears to be a deliberate choice (there's a comment stating that in the code). So just something to live with if we go with this library.

danschwarz:
This is going to need to get squashed before commit, I'm sure.  Too much going on here.
I think html2text may be a good candidate for vendoring. I can put together a version of the library that has the appropriate PRs applied, is Python 3.10 compatible, and formats HTML as markdown "well enough".

PR #360, opened 2023-05-28
If toot activate is invoked without an account, show list of accounts

danschwarz:
```
❯ toot activate
Specify one of the following user accounts to activate:

ds212@babka.social
ds212@blob.cat
ds212@newsie.social
ds212@vivaldi.net
dschwarz@toad.social


~/de/toot activate_user_list *23 ?15 ❯     
```


ihabunek:
Thanks! Merged manually. I just removed the green color because I thought it is a bit much.

PR #374, opened 2023-06-26
Implement data classes for API entities

ihabunek:
Converting JSON returned from the server into data classes makes programming much simpler, especially if your IDE has autocompletion.

Obviously this is a benefit for developers, not the end user, and it costs some extra CPU. In my testing the performance hit seems mostly negligible. It took ~1ms to convert 20 statuses, which is the most allowed by `toot timeline`.

This PR supersedes #313 which was a bit too ambitious. It does not touch TUI code, just uses entities to simplify print functions from `output.py`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
           pytest
       - name: Validate minimum required version
         run: |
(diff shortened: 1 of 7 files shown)


danschwarz (approved):
Looks good to me overall.

PR #377, opened 2023-06-28
Implement a settings file

ihabunek:
Rationale for using tomlkit:

1. Didn't want to use JSON because it's harder to read and write by humans and doesn't support comments, so needed a 3rd party library for another format.
2. TOML seemed like the best candidate, simple and intuitive, doesn't have [gotchas like YAML](https://noyaml.com/)
3. tomlkit seems stable, is recommended in https://docs.python.org/3/library/tomllib.html
4. tomlkit supports writing while preserving style and comments, this will be great once we implement editing settings via CLI or TUI

I'm marking this as experimental because it needs some testing and things may change. Check out the docs page for implemented settings. TUI settings can be next.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -5,6 +5,7 @@
 - [Installation](installation.md)
 - [Usage](usage.md)
     - [Advanced](advanced.md)
+    - [Settings](settings.md)
     - [TUI](tui.md)
 - [Contributing](contributing.md)
     - [Documentation](documentation.md)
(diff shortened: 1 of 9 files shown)


danschwarz:
I agree with the rationale to use TOML for the settings file. 

About the color override, --no-color allows overriding the color setting but only in one direction. Should it be color=yes/no or true/false instead?

danschwarz (approved):
See previous comment, LGTM

PR #378, opened 2023-06-30
Implemented an in-memory cache with LRU eviction for images

danschwarz:
The cache doesn't write to disk, but I think this is a reasonable option. It is set to a 10Mb cache at the moment; we could add a setting and/or command line option to change the cache size.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,4 @@ sphinx
 sphinx-autobuild
 twine
 wheel
+debugpy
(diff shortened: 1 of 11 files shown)


danschwarz:
Thanks to @davesteele for the code I used as a starting point; per his [response to me] (https://gist.github.com/davesteele/44793cd0348f59f8fadd49d7799bd306?permalink_comment_id=4613737#gistcomment-4613737)it is OK to use for GPL2+ compatible licenses.

danschwarz:
closed because it includes unwanted debugger code

PR #380, opened 2023-06-30
Added in-memory image cache with LRU eviction policy

danschwarz:
Hardcoded to 10MB size for now; can add command line and settings options later.

Thanks @davesteele for the GPL2+ licensed code I used as a starting point


danschwarz:
AAARGH.

Issue #370, opened 2023-06-22
Support for hashtag timelines in TUI


andypiper:
I'd love to be able to check my followed hashtags in the TUI.

danschwarz:
You can! It's just a bit non-obvious how to get there.

[G]o to the Go To menu.  Scroll down to Hashtag: and enter in your hashtag of choice, then select Local or Public timeline.
![image](https://github.com/ihabunek/toot/assets/3261094/58616357-b1e9-43b6-a7ff-29884717d46e)

Now you're viewing the hashtag timeline.
![image](https://github.com/ihabunek/toot/assets/3261094/b8ba222a-21fa-4a75-a543-50ad6bd501ef)

Now you should be able to Pin that timeline with key [E] so it shows up under Shortcuts in the Go To menu in future.

![image](https://github.com/ihabunek/toot/assets/3261094/a1cf6536-215e-418e-a244-1138e6b9e546)

I just realized the help text lists [P] for Pin rather than E 🤷 . I'll fix that.

andypiper:
Thanks, this was indeed a bit unclear but you've helped me to understand it.

In Mastodon v4+ servers, a user can have a list of followed hashtags on the server / stored in the profile, it would be nice to be able to select one of these and pin in toot.

`curl --header "Authorization: Bearer $TOKEN" https://macaw.social/api/v1/followed_tags | jq '.[] | .name'`

PR #382, opened 2023-07-02
Changed [H]elp key to [?], Correctly documented Sav[e] key as E in help overlay

### LINKED ISSUE #280: H key in TUI is advertised as both cursor movement + Help; can't do both.
H as move-left doesn't have much use now, but if pull request #277  is merged, move-left and move-right will be useful to move focus from timeline list to status detail and back again.

Suggestion: change the TUI to use '?' for help instead

image

danschwarz:
Fixes #280 , #370


ihabunek:
Merged manually, thanks.

PR #383, opened 2023-07-06
Changed [L]inks key to L[i]nks so that L key is CURSOR RIGHT

danschwarz:
Now H J K L are all free to do cursor movement. Whether this is worth it or not, I don't know?


ihabunek (approved):
Thanks!

PR #384, opened 2023-07-07
Added sanity check tests to ensure PALETTE and MONO_PALETTE match

danschwarz:
And fixed a bug uncovered by the above tests :)


ihabunek (approved):
Thanks, good idea.

PR #385, opened 2023-07-08
Read TUI colors from settings

ihabunek:
After playing around with Urwid colors I decided to remove the monochrome palette since it's possible to define the monochrome foreground color as the third color in the color tuple. Docs here: http://urwid.org/tutorial/index.html#display-attributes

It's not possible to change the background in monochrome mode, so the black-on-white colors didn't work for me at all.


danschwarz (commented):
Use of the monochrome foreground color is possible, but, as you note, you lose the ability to specify black text on a white cell background.  I thought the use of black text on a white background makes the TUI look somewhat better, and it also allows you to properly display poll bars.

The tradeoff, of course, is a bit more hassle maintaining the two palette lists.

The choice is yours, I'm fine with either approach. Poll bars could be rewritten to use Unicode U+2588 full block characters if we go with your approach.

ihabunek:
That's interesting. I was testing in kitty, and in monochorme mode, I did not get black-on-white text at all. It works in gnome terminal though.

Kitty left, gnome term right:

image

I'll probably add the monochrome palette back.

danschwarz:
I can also reproduce the behavior with Windows Terminal, Kitty, and Xterm.js.

ihabunek:
I'm merging this for now so I can release, will think about improvements later.

Issue #386, opened 2023-07-13
Allow to search/target original URL of a post


nemobis:
If I understand correctly, all commands require a local ID for posts. Using the original URL anywhere will fail, including in search.

```
$ toot bookmark https://mastodon.ar.al/@aral/100666978784763896
Unknown error
$ toot search https://mastodon.ar.al/@aral/100666978784763896
Nothing found
```

If I manually search that URL on the web UI, I get https://mamot.fr/@aral@mastodon.ar.al/100666978842964982 and then I can use the local ID:

```
$ toot bookmark 100666978842964982
✓ Status bookmarked
```

It would be nice to be able to use the original URL and let toot do the search (so that the post is identified, or cached first if it wasn't, and the local ID returned).

Tested on mamot.fr (v4.1.3)

nemobis:
When using the v2 search API , the local ID is in the `id` field inside `statuses`.

For example from https://mamot.fr/api/v2/search?q=https:%2F%2Fmastodon.social%2Fusers%2Fhhardy01%2Fstatuses%2F99739317870899852&resolve=true&limit=5 I get:

```
{"accounts":[],"statuses":[{"id":"99739317840845686","created_at":"2018-03-24T14:20:12.000Z",}]}
```

(JSON trimmed) for https://mamot.fr/@hhardy01@mastodon.social/99739317840845686.

nemobis:
Any chance of getting the PR merged? It seems relatively straightforward to me but if changes are needed I can try. I still need this feature and I'd rather not maintain a fork just for it. :)

dbdowjfb:
As of v0.40.2, toot seems to support `toot search --json`. So it can now output json for the search result

ihabunek:
Available in 0.44.1. Thanks everyone and sorry for the delay.

I'm a little worried that searching for toots by URL is not 100% reliable, so I'm leaving this undocumented for now. I'd appreciate if you would test and verify everything is working as expected.

PR #388, opened 2023-07-15
Support bookmarking a status by URL

nemobis:
Addresses issue #386 but not in every case.


ihabunek:
Thanks for the contribution. I've had a lot of work with the big changes released in 0.40.0 so didn't have time to review and test this. I'll get back to you soon since this is an interesting problem...

PR #389, opened 2023-07-20
Added stub term-image library for use when term-image can't load

danschwarz:
Also, conditional code throughout to detect the situation where term-image can't load. We can't actually call stub library functions so this is necessary.

The stub library is based on the term-image 0.7.0 / 0.8.0 API; it will need to change as the API changes.  API should stabilize when term-image hits 1.0.0.

This completes the work @ihabunek started, to allow the TUI to work properly when term-image is not available.

Obviously, images do not display, but the app runs without crashing and visually the TUI is similar (but not the same) as toot tui without image support.

This PR also adds a global `has_term_image` to `timeline.py`, `overlays.py`, and `utils.py`, along with checks as necessary, to avoid instantiating classes or calling functions in the stub library.

This feels like a kludge, but it works. If you know of a better way to do this, let me know.


ihabunek:
@danschwarz Just to let you know that I'm travelling next week, and moving house when i get back. So unfortunately don't expect much from me until September. Sorry about that.

danschwarz:
Thanks for letting me know. Good luck with the move.

danschwarz:
Working on something better & cleaner to replace this

Issue #392, opened 2023-08-01
'm' keypress on timeline with media attachment pops up ImageMagick logo


draeath:
A picture is worth a thousand words, so here's what it looks like. I can't really tell if this is a problem with my environment (ImageMagick has *really stupid binary names*) or toot itself.

![what](https://github.com/ihabunek/toot/assets/987449/659154bb-3ea5-4d05-9860-e6b9a27beb70)

To replicate:

1. Launch `toot tui`
2. hit `g` and select "Global public timeline."
3. Scroll to any post that contains an media attachment URL
4. Press `m`

My brain skipped a few steps, for some reason I expected this to pop up the 'mute' that we see when we first press `a` to see the poster's account information.

I have absolutely no idea why IMageMagick is popping up, and this key is not documented in the TUI's help screen or in the status bar. Also note the error in the status bar:

> display: attempt to perform an operation not allowed by the security policy `HTTPS' @ error/constitute.c/IsCoderAuthorized/454.

----

Some environmental details:

- OpenSUSE tumbleweed (snapshot 20230730-0)
- toot v0.38.1 installed via pip, within a python 3.11 venv
- toot is running inside konsole via an X.org session

package ImageMagick-7.1.1.12-1.1.x86_64 installed, which provides these binaries in my path:
- /usr/bin/animate
- /usr/bin/compare
- /usr/bin/composite
- /usr/bin/conjure
- /usr/bin/convert
- /usr/bin/display
- /usr/bin/identify
- /usr/bin/import
- /usr/bin/magick
- /usr/bin/magick-script
- /usr/bin/mogrify
- /usr/bin/montage
- /usr/bin/stream

draeath:
Putting the viewer detection itself aside, this exception should probably be handled with a status line message instead of being unhandled.

draeath:
I suggest:

1. remove or deprioritize `display` from the list, since ImageMagick seems it doesn't understand what to do with a URL
2. add `gwenview` to cover KDE users, it fills the same niche as `eog` does for Gnome. If given a URL as an argument, it does the correct thing and displays the image correctly. (tested with gwenview 23.04.3)

draeath:
Looks like `display` is at the end of the list.

Submitted PR #393 to add gwenview between it and `eog`

PR #393, opened 2023-08-01
add gwenview image viewer for TUI

draeath:
add `gwenview` as noted in #392


ihabunek:
Thanks! The latest release lets you configure the media viewer via settings. See https://toot.bezdomni.net/settings.html#tui-view-images

PR #403, opened 2023-09-22
Support to display a limited set of HTML tags (aligned with mastodon 4.2) plus URL support (soft dependency on urwidgets library)

danschwarz:
This is essentially a rewrite of the closed PR #348 using @AnonymouX47 's `TextEmbed` widget from `urwidgets`. It does not suffer from the rendering issue that caused problems for the previous PR, that used `urwid.Text`.   As a bonus, it includes support for clickable links on terminals that support OCS 8 hyperlinks.

If `urwidgets` library is not available, we fall back seamlessly to using urwid.Text and everything works as normal, you just don't get clickable links on terminals that support OCS 8 hyperlinks.

This PR renders all HTML elements, and some HTML attributes, that are supported in Mastodon 4.2 [spec](https://docs.joinmastodon.org/spec/activitypub/#sanitization)

### HTML Status on a glitch-soc server
image

### The same status on the TUI with this PR
image

As the `urwidgets` dependency is Python 3.7+,  I have removed the Python 3.6 automated Github tests.


danschwarz:
"workflow runs completed with no jobs"
🤔

danschwarz:
See new simplified PR with all changes squashed into a single commit #404

Issue #399, opened 2023-08-30
toot timeline chokes on last_status_at field


illdefined:
When retrieving the timeline from an Akkoma instance (tried it with Akkoma 3.9.3, 3.10.3 and 3.10.4), `toot timeline` (version 0.38.1) chokes on the `last_status_at` account field:

```
Traceback (most recent call last):
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/bin/.toot-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/console.py", line 961, in main
    run_command(app, user, command_name, args)
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/console.py", line 943, in run_command
    return fn(app, user, parsed_args)
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/commands.py", line 60, in timeline
    statuses = [from_dict(Status, item) for item in items]
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/commands.py", line 60, in <listcomp>
    statuses = [from_dict(Status, item) for item in items]
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/entities.py", line 373, in from_dict
    return cls(**dict(_fields()))
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/entities.py", line 371, in _fields
    yield field.name, _convert(field_type, value)
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/entities.py", line 404, in _convert
    return from_dict(field_type, value)
  File "/nix/store/85x1fp77svwvdmbnfpzvn63qpg0vlzbj-toot-0.38.1/lib/python3.10/site-packages/toot/entities.py", line 373, in from_dict
    return cls(**dict(_fields()))
[...]

PR #405, opened 2023-09-23
Fix crash when parsing date

snan:
I was getting a `ValueError: Invalid isoformat string:
'2023-09-23T13:07:50'` crash.

### DIFF
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -122,6 +122,7 @@ def test_timeline(mock_get, monkeypatch, capsys):
         'id': '111111111111111111',
         'account': {
             'display_name': 'Frank Zappa 🎸',
+            'last_status_at': '2017-04-12T15:53:18.174Z',
             'acct': 'fz'
         },
         'created_at': '2017-04-12T15:53:18.174Z',
@@ -164,6 +165,7 @@ def test_timeline_with_re(mock_get, monkeypatch, capsys):
             'created_at': '2017-04-12T15:53:18.174Z',
             'account': {
                 'display_name': 'Johnny Cash',
+                'last_status_at': '2011-04-12',
                 'acct': 'jc'
             },
             'content': "<p>The computer can&apos;t tell you the emotional story. It can give you the exact mathematical design, but what's missing is the eyebrows.</p>",
--- a/toot/entities.py
+++ b/toot/entities.py
@@ -394,7 +394,7 @@ def _convert(field_type, value):
         return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
 
     if field_type == date:
-        return date.fromisoformat(value)
+        return datetime.fromisoformat(value).date()
 
     if get_origin(field_type) == list:
         (inner_type,) = get_args(field_type)
--- a/toot/entities.py
+++ b/toot/entities.py
@@ -394,7 +394,7 @@ def _convert(field_type, value):
         return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
 
     if field_type == date:
-        return datetime.fromisoformat(value).date()
+        return datetime.fromisoformat(value.replace('Z', '+00:00')).date()
 
     if get_origin(field_type) == list:
         (inner_type,) = get_args(field_type)


snan:
Dangit, I didn't realize dateutils wasn't part of the standard library 🤦🏻‍♀️

danschwarz:
Can you fix this without introducing a new library dependency?  The datetime string you posted, `2023-09-23T13:07:50`, looks correct to me - what's going on with the existing code?

snan:
> Thanks for the PR. Can you fix this without introducing a new library dependency? The datetime string you posted, `2023-09-23T13:07:50`, looks correct to me - what's going on with the existing code?

When running `toot notifications -rm` on master, it crashes with this error message:

    ValueError: Invalid isoformat string: '2023-09-23T13:07:50'

I can verify that it does by running `python3` as a REPL and pasting in:

    from datetime import date, datetime
    date.fromisoformat('2023-09-23T13:07:50')

That crashes with the same error message for me. I then found [this StackExchange thread](https://stackoverflow.com/questions/55542280/why-does-python-3-find-this-iso8601-date-2019-04-05t165526z-invalid) which is what suggested dateutil 🤷🏻‍♀️

snan:
I'm on Debian GNU/Linux 12 (bookworm). The Debian stable version of toot (which is 0.34) doesn't have this issue but I downloaded the git master version to work on something else and then I ran into this.

According to [that aforementioned thread](https://stackoverflow.com/questions/55542280/why-does-python-3-find-this-iso8601-date-2019-04-05t165526z-invalid) support for this was [added in version 3.11](https://docs.python.org/3.11/library/datetime.html#datetime.datetime.fromisoformat) so people on that version won't be able to replicate the bug. Still a problem for me though 🤷🏻‍♀️

danschwarz:
OK thanks for the explanation, seems reasonable to me to add python-dateutil as we have a lot of date parsing that we need to get right. Hopefully @ihabunek will be back soon to weigh in on this. Meanwhile, if you can fix up the imports so the automated tests pass, I think this PR will be acccepted.

snan:
Not sure how to do that 💁🏻‍♀️

danschwarz:
[Here's](https://discuss.python.org/t/parse-z-timezone-suffix-in-datetime/2220) an interesting workaround that doesn't require us to pull in datetime. From the thread:

`datetime.fromisoformat(my_date.replace('Z', '+00:00'))`

if i run `python3` as a REPL, I first verified that the exception gets thrown when using `date.fromisoformat` by itself.   But:

```
>>> datetime.fromisoformat('2023-09-23T13:07:50'.replace('Z', '+00:00'))
datetime.datetime(2023, 9, 23, 13, 7, 50)
```
works as expected, no exception thrown.  Granted, `dateutil` is more powerful and flexible, but this seems to do the job. And it should have no problem passing the automated tests.

danschwarz:
Simpler fix for all of this:

rather than calling:
`date.fromisoformat('2023-09-23T13:07:50')`
call:
`datetime.fromisoformat('2023-09-23T13:07:50')`
then convert the datetime back to a date with `datetime.date()`

And, for good measure, convert "Z" suffix to +00:00 as per [this thread](https://discuss.python.org/t/parse-z-timezone-suffix-in-datetime/2220) 

Running that as a REPL with python3 (I'm running 3.10, which doesn't have the 3.11 improvements to `fromisoformat`)
```
>>> datetime.fromisoformat('2023-09-23T13:07:50'.replace('Z', '+00:00')).date()
datetime.date(2023, 9, 23)
```
This removes the need to import python-dateutil (granted, it's more robust, but we can make that change later if necessary.).  And the code change should pass the automated tests.

snan:
> datetime.fromisoformat('2023-09-23T13:07:50')

I can verify that that doesn't crash ♥♥👍🏻👍🏻

danschwarz:
More interesting things...

#399 notes a documentation discrepancy in Mastodon regarding this field, and the toot crash

re: "controversial timezone workaround" - this was controversial _at the time_, but it seems that those who wanted to expand `datetime.fromisoformat` functionality won the argument, and starting in Python 3.11, it handles more variations of ISO 8601 datetime formats. 🤷‍♂️ 

Alternately, we could just use `datetime.strptime` as the datetime handling code a few lines above does, and forget about `datetime.fromisoformat` entirely 😬 

`datetime.strptime(value[:10], '%Y-%m-%d').date()`

By the way - There was no test case exercising this code path in `entities.py`; to fix that we can  update `test_console.py` with added last_status_at fields in two formats: ISO 8601 date, and ISO 8601 timestamp.  So this should cover the two variations we know of right now.

Here's a patch file that adds the fields to the test cases. If you would git patch your branch with this file and push the changes to the PR, I'd appreciate it.

[tc.patch](https://github.com/ihabunek/toot/files/12707541/tc.patch)

snan:
Thank you for sending a patch 👍🏻 (instead of a PR on my branch 😵‍💫)

danschwarz (approved):

PR #406, opened 2023-09-23
Clarify misleading venv docs

snan:
On my zsh I was still getting /usr/bin/toot even in the venv.


danschwarz:
This is not my experience. after executing `. venv/bin/activate`,
```
❯ which toot
/home/dschwarz/dev/toot/venv/bin/toot
```
as expected.

Not sure what's going on in your environment?

snan:
I use a shell where path changes aren't applied until next start. Or something, IDK.

ihabunek:
Thanks.

Issue #416, opened 2023-11-18
feature request: show markdown messages in markdown, not plain text


lilmike:
Hiya,

I notice that when someone I follow posts a message using markdown, toot shows just the converted text content, which leaves out certain important things, such as links which simply show the text of the link, but not the link itself. Perhaps if the toot is written in markdown, the raw markdown should be displayed. This is using pleroma, I don't know if it's different on mastodon.

-Michael.

danschwarz:
This gets tricky for two reasons:

1. I think, but need to verify, that Pleroma is using its own field, pleroma/content, alongside standard content in Status messages. Toot doesn't have special code to detect and show the pleroma/content field; it's just showing the regular content field.  Could this be changed? Yes.  _Should_ it be changed for benefit of Pleroma created content? Dunno.
2. Displaying just the text of a link, but not the link itself - I assume you mean showing the text as a clickable link like [this](http://www.example.com).  Well, that can be done, if and only if your terminal emulator supports the [OSC 8 standard](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).  We're adding support for that now in Master branch, with PR #415 just merged.  But there's no way for Toot to detect OCS 8 support itself, so we only use it in a limited fashion, by making existing http(s):// links clickable.

lilmike:
Sorry, I wasn't super clear there: What I meant was, if you have
markdown code like this:

```
[this is a link](http://example.com)
```

Then toot will simply show 'this is a link' and the url will be nowhere
to be found.

-Michael.

ihabunek:
I just tested and pleroma does not return the markdown representation of a toot even if it was posted in markdown.

The toot contained: `test links [bezdomni](https://bezdomni.net)`

Below is the json that Pleroma returns for the posted status. Notice that `content` is in HTML and `pleroma.content.text/plain` is plaintext. There is no markdown version.

I presume they convert the markdown (or bbcode which is also supported) into HTML and that's the canonical representation.

We could use something like [html2text](https://github.com/Alir3z4/html2text/) to convert back to markdown and show that when not using richtext.

```json
{
    "account": { ... },
    "application": null,
    "bookmarked": false,
    "card": null,
    "content": "<p>test links <a href=\"https://bezdomni.net\">bezdomni</a></p>",
    "created_at": "2023-11-19T11:18:51.000Z",
    "edited_at": null,
    "emojis": [],
    "favourited": false,
    "favourites_count": 0,
[...]

PR #417, opened 2023-11-21
Proof of concept for issue #416 - adds Markdown rendering in TUI

danschwarz:
Here's a PR against master that uses [html2text](https://pypi.org/project/html2text/) to render statuses in Markdown. (per issue #416)  It does this all the time for console output (`toot timeline` etc.) and copy-to-clipboard.  In the TUI, it renders Markdown as a fallback if the `urwidgets` library is unavailable.

It works OK, but has some rendering challenges due to quirks and limitations of html2text.  I think I'll try substituting ~~[markdown-it-py](https://github.com/executablebooks/markdown-it-py) to see if it does a better job.~~  Turns out, markdown-it-py goes the _opposite direction_, from markdown to HTML.

I tested [markdownify](https://github.com/matthewwithanm/python-markdownify) instead; it's less flexible than html2text and has its own quirks, so it's a question of which library's quirks we can put up with.  Or find another.  The only other solution I've found is [pandoc](https://pypi.org/project/pandoc/),  which relies on a Haskell library. So that's not ideal.

For the moment, I'm sticking with html2text.

image

Another issue is that urwid Text widgets wrap text as needed, and html2text depends on fixed width lines in order to render things like blockquotes and lists correctly.  If the TUI status window is too narrow, blockquotes and lists won't look right.  Not much we can do about this other than detect the status window width and re-render the Markdown on any window resize 🤮 .  Or ignore the problem (my preferred solution.)   You can see an example of this line wrapping problem in the image below:
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
         "beautifulsoup4>=4.5.0,<5.0",
         "wcwidth>=0.1.7",
         "urwid>=2.0.0,<3.0",
-        "tomlkit>=0.10.0,<1.0"
+        "tomlkit>=0.10.0,<1.0",
         "html2text>=2020.1.16"
     ],
     extras_require={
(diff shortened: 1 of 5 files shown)


danschwarz:
Opening a new PR that uses Pandoc instead (turns out, it's better)

PR #425, opened 2023-11-26
Add support for markdown in TUI, console, clipboard using pypandoc library

danschwarz:
As it turns out, the `pandoc` utility, with `pypandoc` Python bindings, do a better job converting HTML to markdown than `html2text` or `markdownify` libraries. `Pandoc` is widely used and maintained, in contrast to the other libraries.

In the TUI, statuses are rendered in HTML by `urwidgets` if available, fallback is markdown (via `pandoc` if available), and finally plaintext if neither `urwidgets` nor `pandoc` are available.

In the console and clipboard, status rendering is in markdown via `pandoc`, fallback to plaintext if `pandoc` is not available.

TODO:
1. Command line option or config file option to choose tui status rendering: HTML, markdown, or plaintext.
2. Command line option or config file option to choose console status rendering: markdown or plaintext.
3. Test cases to exercise all three rendering options correctly. At present, test_console.py (correctly) complains because statuses rendered as markdown do not match the expected plaintext rendering.

Here's an example of markdown rendering in the TUI.  This is using github-flavored-markdown, FWIW.

This addresses issue #416 .

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,8 @@ setup(
         "wcwidth>=0.1.7",
         "urwid>=2.0.0,<3.0",
         "tomlkit>=0.10.0,<1.0",
-        "html2text>=2020.1.16"
+        "pypandoc>=1.12.0,<2.0",
+        "pypandoc-binary>=1.12.0,<2.0"
     ],
     extras_require={
         # Required to display rich text in the TUI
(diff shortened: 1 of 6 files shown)


danschwarz:
something is wrong with the code in this PR. fixing.

PR #427, opened 2023-11-26
Add support for Markdown in TUI, console, clipboard output

danschwarz:
As it turns out, the `pandoc` utility, with `pypandoc` Python bindings, do a better job converting HTML to markdown than `html2text` or `markdownify` libraries. `Pandoc` is widely used and maintained, in contrast to the other libraries.

In the TUI, statuses are rendered in HTML by `urwidgets` if available, fallback is markdown (via `pandoc` if available), and finally plaintext if neither `urwidgets` nor `pandoc` are available.

In the console and clipboard, status rendering is in markdown via `pandoc`, fallback to plaintext if `pandoc` is not available.

TODO: 

- [x] Command line option or config file option to choose tui status rendering: HTML, markdown, or plaintext.
- [X] Command line option or config file option to choose console status rendering: markdown or plaintext.
- [ ] Test cases to exercise all rendering options correctly. At present, we are testing only the plaintext rendering in console.

Here's an example of markdown rendering in the TUI.  This is using github-flavored-markdown, FWIW.

This addresses issue #416 .

image


danschwarz:
Switched it to default to plaintext (as it was originally) and use the -md / --markdown option to render as markdown. Makes more sense this way.

ihabunek:
Sorry this never made it in, Dan.

PR #428, opened 2023-11-29
Migrate to Click

### LINKED ISSUE #375: Migrate from argparse to click
[Click](https://click.palletsprojects.com/en/8.1.x/options/) is a common alternative to argparse. And while I'm not a big fan of the annotations-based workflow, it would make several features trivial to implement, which are a pain using argparse. Most notably:

1. Passing default values [from environment variables](https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables)
2. Passing default values [from a config file](https://click.palletsprojects.com/en/8.1.x/commands/#overriding-defaults)
3. [Shell completion](https://click.palletsprojects.com/en/8.1.x/shell-completion/)

For their own list of benefits see:
https://click.palletsprojects.com/en/8.1.x/why/

There are also several contrib projects which sound like we'd want to use them, e.g. generate man pages:
https://github.com/click-contrib/

The downside is that it might be a bit slower than the current implementation. Currently we don't create the argparse definitions for all commands on each run, but rather only for the one command which is being executed. This is something which should be tested before commiting down this path.

The other downside is a pretty big rewrite of `console.py` which would require some work.

ihabunek:
Rewrite toot to use Click. 

TODO before merging:
* implement `--quiet` flag, currently does nothing
* add decorator for commands which require authentication, print an error when not authenticated
* settings: deprecate `[tui]` category in settings in favour of `[commands.tui]`
* optional: settings: deprecate `[commands.*]` sections in favour of `[toot.*]`?

fixes #375


ihabunek:
I merged this now to avoid having a long running branch. The few changes which need to be implemented before release can be done on master.

PR #437, opened 2023-12-06
Add install make target

### LINKED ISSUE #433: [Feature Request] Provide an "install" Makefile target
I was happy that `make` worked, but disappointed not to find a `make install`.
The good thing about standards (these are covered by the GNU Coding Standards) is that you don't need to know the details of a specific package to build and install it, would be nice if this was the case here too.

strk:
Closes #433


strk:
Interesting position from Debian-12:
```
pip install .
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

ihabunek:
The problem with pip install is that it installs the package globally and that can cause conflicts with packages installed via package manager. 

This is why they added externally-managed-environment seen in Debian 12. This will presumably be adopted by most distros.
See https://peps.python.org/pep-0668/

You could do pip install --user but usually the best way is to create a virtual env.

Python packaging is a bit of a mess.

-- Ivan

On Thu, 7 Dec 2023, at 12:26, Sandro Santilli wrote:
> 
> 
> Interesting position from Debian-12:
> 
> `pip install .
> error: externally-managed-environment
> 
> × This environment is externally managed
> ╰─> To install Python packages system-wide, try apt install
>     python3-xyz, where xyz is the package you are trying to
>     install.
>     
>     If you wish to install a non-Debian-packaged Python package,
>     create a virtual environment using python3 -m venv path/to/venv.
>     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
>     sure you have python3-full installed.
>     
>     If you wish to install a non-Debian packaged Python application,
>     it may be easiest to use pipx install xyz, which will manage a
>     virtual environment for you. Make sure you have pipx installed.
>     
>     See /usr/share/doc/python3.11/README.venv for more information.
> 
> note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
> hint: See PEP 668 for the detailed specification.
`

strk:
I've found `pipx` being recommended by some python developers (and even by https://peps.python.org/pep-0668/), which is like pip but deals with virtual envs.

It also allows using a system-wide virtualenv, using a pair of enviromnent variables:
https://github.com/pypa/pipx/issues/754#issuecomment-951162846

How would the maintainer feel about going in that direction and requiring `pipx` for install ?

The install target would then be something like:

> PIPX_HOME=$(pipx_home) PIPX_BIN_DIR=$(bindir) pipx install .

What to make of pipx_home default I'm not sure, maybe $(sharedir)/pipx ?

strk:
I've pushed a commit using pipx

ihabunek:
Thanks for the pull, but I decided not to merge this one.

It makes assumptions like that the user wants to use pipx, that pipx home defaults to `/opt/pipx`, etc.

If you use toot from source, some python knowledge is required, and every user can choose what suits their needs.

For example, I never use pipx to install from source, i install using pip into a virtual env managed by virtualenvwrapper. But that's just my workflow.

Issue #422, opened 2023-11-25
toot timeline gets number of terminal lines wrong


strk:
typing `toot timeline` dumps many posts/toots to the terminal and at the end asks me if I want to continue, but not all the posts printed are visible in my terminal and I have to scroll the terminal backlog to see them all. I suppose that's not intentional and is the effect of a wrong detection of the number of lines in the terminal OR of the number of lines in the posts

danschwarz:
It's currently written to pause after `--count` status messages (default: 10) without regard to screen rows.  I agree it's annoying and should default to displaying MAX(1, # of statuses that fit on a screen) per prompt.  How much of a hassle to implement? I'll have a look 🤷‍♂️

PR #439, opened 2023-12-06
added --color flag to force color even if stdout is a tty

danschwarz:
This option was inspired by @strk comment:
https://github.com/ihabunek/toot/issues/422#issuecomment-1842649257

Now there is a `--color` option that works exactly like (but opposite to) the `--no-color` option. 

There are other ways to implement this, possibly by replacing `--no-color` with something like `--color` with values like `yes`, `no`, `auto`  (auto being default and what the app currently does if no flag passed).

If you'd rather see the flag implemented that way, I can change the code.  Otherwise, this is good to go and allows things like

`toot timeline --count 20 --color | less -R`

to display timeline output, properly paged, in color.

### DIFF
--- a/toot/console.py
+++ b/toot/console.py
@@ -164,6 +164,11 @@ common_args = [
         "action": 'store_true',
         "default": False,
     }),
+    (["--color"], {
+        "help": "always use ANSI colors in output",
+        "action": 'store_true',
+        "default": False,
+    }),
     (["--quiet"], {
         "help": "don't write to stdout on success",
         "action": 'store_true',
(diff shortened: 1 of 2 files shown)


ihabunek:
Thanks, but I'm in the middle of the Click rewrite so it will be handled there. See #428, work in progress.

strk:
On December 7, 2023 9:40:10 AM GMT+01:00, Ivan Habunek ***@***.***> wrote:
>Thanks, but I'm in the middle of the Click rewrite so it will be handled there. See #428, work in progress.
>

Do you accept pull requests for the 'click' branch ?
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ihabunek:
The code there is very much in flux so i would prefer if you waited until it is merged. What did you have in mind?

-- Ivan

strk:
> What did you have in mind?

Nothing in particular, just this PR itself

PR #444, opened 2023-12-14
Add install-bash-completion and install-shell-completion targets

strk:
I've only done bash as I use bash and have no idea how zsh and fish deal with custom completion files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,3 +46,15 @@ bundle:
 		--output toot-`git describe`.pyz bundle \
 		--compress
 	echo "Bundle created: toot-`git describe`.pyz"
+
+
+# TODO: add more rules for more shells
+.PHONY: install-shell-completion
+install-shell-completion: install-bash-completion
+	@echo "See docs/shell_completion.md for Fish and Zsh completions"
+
+.PHONY: install-bash-completion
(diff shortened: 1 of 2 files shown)


strk:
The completion seems to be not working with the click branch, will check if it works with the master branch

ihabunek:
It's only implemented on the click branch. Not yet merged to master. It works for me on zsh.

strk:
For bash it fails:
```
eval "$(_TOOT_COMPLETE=bash_source toot)"
Unknown command '-'
Run toot --help to show a list of available commands.
v0.39.0: command not found
Authentication:: command not found
bash: syntax error near unexpected token `('
```

Is this something supposed to be implemented by click itself ?

strk:
It looks like _TOOT_COMPLETE env variable has no effects at all on toot:
```
$ _TOOT_COMPLETE=bash_source toot
toot - a Mastodon CLI client
v0.39.0

Authentication:
  toot login            Log into a mastodon instance using your browser (recommended)
...
```

I guess it would be a lot safer to just build the completion scriptsi on `make` and install the generated output, rather than calling the dangerous `eval` on every new shell

strk:
this PR continues in #446

PR #445, opened 2023-12-14
Make bundle rule phony

strk:
So it is never considered "done"


strk:
Makes it fail on second invocation but I prefer a failure over a fake sense of everything being correctly bundled (isn't this what happens now?)

ihabunek:
Thanks

PR #446, opened 2023-12-14
Add install-bash-completion and install-shell-completion targets

strk:
To later be hooked to the generic 'install' rule


ihabunek:
Again, I'd like to keep the makefile dead simple, it's more a thing for my convenience than anything.

PR #447, opened 2023-12-19
Updated min version of urwid to 2.1.0 aligned with urwidgets library

danschwarz:
Urwidgets library requires urwid 2.1.0 so I've adjusted our minimum urwid version acordingly.

Note that urwidgets is an optional dependency, so if we really want to, we can leave our urwid version at 2.0.0.


ihabunek:
Thanks, but I would prefer to leave it at 2.0.0 since repos for some older distro versions still carry urwid 2.0.1.
https://repology.org/project/urwid/versions

PR #450, opened 2023-12-31
add an option to automatically expand content warnings

lexiwinter:
- Add an option to expand CWs automatically.
- Display the CW notice even on expanded toot, so it's still there even when this option is enabled.


ihabunek:
Thanks.

PR #451, opened 2023-12-31
tui: allow editing toots

lexiwinter:
Add new [E]dit command to the timeline: opens an existing toot to allow editing it.  Since this is more or less the same operation as posting a new toot, extend the StatusComposer view to support this rather than implementing a new view.

---

~~this is a draft PR because of a significant unsolved issue: the Mastodon API returns toots as HTML, not plain text, so we would need to strip the HTML from the toot before displaying it in StatusComposer.  this is particularly difficult for toots with links or embeds.  i'm wondering if the API provides a better way to do this, but i couldn't find anything obvious in the docs.~~

~~aside from that, not all of the toot options are copied to the edited toot right now (i'll fix this)~~, and it could also do with more testing.

~~also, the edited toot currently doesn't show up on the timeline without a refresh.~~

### DIFF
--- a/toot/api.py
+++ b/toot/api.py
@@ -230,12 +230,67 @@ def post_status(
     return http.post(app, user, '/api/v1/statuses', json=data, headers=headers)
 
 
+def edit_status(
+    app,
+    user,
+    id,
+    status,
+    visibility='public',
+    media_ids=None,
+    sensitive=False,
+    spoiler_text=None,
+    in_reply_to_id=None,
+    language=None,
+    content_type=None,
+    poll_options=None,
+    poll_expires_in=None,
+    poll_multiple=None,
+    poll_hide_totals=None,
+) -> Response:
+    """
+    Edit an existing status
+    https://docs.joinmastodon.org/methods/statuses/#edit
+    """
+
+    # Strip keys for which value is None
+    # Sending null values doesn't bother Mastodon, but it breaks Pleroma
+    data = drop_empty_values({
+        'status': status,
+        'media_ids': media_ids,
+        'visibility': visibility,
+        'sensitive': sensitive,
+        'in_reply_to_id': in_reply_to_id,
+        'language': language,
+        'content_type': content_type,
+        'spoiler_text': spoiler_text,
+    })
+
+    if poll_options:
+        data["poll"] = {
+            "options": poll_options,
+            "expires_in": poll_expires_in,
+            "multiple": poll_multiple,
+            "hide_totals": poll_hide_totals,
+        }
+
+    return http.put(app, user, f"/api/v1/statuses/{id}", json=data)
+
+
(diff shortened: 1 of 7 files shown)


ihabunek:
Thanks! This requires more brain cells than I'm willing to invest today. I'll leave the review for next year. :)

danschwarz:
I suggest we don't strip the HTML when editing the toot.  Rather, convert it to markdown, edit, then convert back to HTML.

#427 gives the required support libraries to reliably do HTML>markdown conversion (if you're willing to put up with the pandoc+pypandoc dependencies, it really does a better job than the alternatives.)  I think those libraries will do markdown>HTML conversion as well, but will have to check.

lexiwinter:
i had a look at another library, Mastodon.py, and found [it implements an endpoint](https://mastodonpy.readthedocs.io/en/1.8.1/_modules/mastodon/statuses.html#Mastodon.status_source) called `/api/v1/statuses/:id/source` to fetch the plain-text source of a toot. it turns out [this actually is documented in the Mastodon API docs](https://docs.joinmastodon.org/methods/statuses/#source), i just missed it the first time i looked.

according to [a brief note](https://docs.joinmastodon.org/api/guidelines/#formatting), this only works for local toots, but only local toots can be edited, so this is fine.

hopefully, this means we don't need to do any conversion ourselves or use Pandoc, although i only use Mastodon so i don't know how this works on other implementations; presumably implementations that support Markdown will return the Markdown source.

ihabunek:
Good find! Was not aware of that. It's been a while since I read through the whole API docs to see what's new.

lexiwinter:
- rebased against current master
- now uses the /source API method to fetch the original text
- original content warning is displayed in the edit box
- polls currently aren't copied over, so will be removed on edit
- fix an unrelated flake8 error in `toot/cli/__init__.py`

unless there are any other comments, the only thing left to land this is support for editing polls.

danschwarz:
As a first implementation this would be fine; we can't author rich text statuses in toot yet anyway.

lexiwinter:
fixed an issue where editing a toot *without* a poll would fail (because the 'poll' key is still in the status dict, it's just set to null).

lexiwinter:
i've just noticed the issue where the updated status content isn't shown until refresh is still there.  i'll add something to fix that, it looks fairly straightforward.

lexiwinter:
fixed.

ihabunek (approved):
Looks good, tested it a bit and seems to work fine.

danschwarz:
Interesting discovery- when editing toots containing rich text (html) on a glitch-doc server, the API call you're using automatically translates the html into markdown server-side. And reverses the translation when you submit the edit. At least this was my experience. I'll do some more testing. Unexpected and nice benefit.

PR #452, opened 2023-12-31
tui: fix display glitch for reply icon in timeline

lexiwinter:
In some fonts, "⤶" (U+2936 ARROW POINTING DOWNWARDS THEN CURVING LEFTWARDS) may be a double-width character.  To avoid a display glitch where this overlaps with the boosted icon, print a space after it.


ihabunek:
Thanks. I'm actually thinking whether using symbols here is more trouble than it's worth and we should switch over to something more ASCII.

PR #453, opened 2023-12-31
tui: show edit date in toot view

lexiwinter:
When viewing a toot which has been edited, show the edit date.

While here, fix a bug where the '*' edit marker in the timeline wouldn't show for retoots because it was checking the retoot status instead of the original status.


ihabunek (approved):

ihabunek:
Thanks!

PR #454, opened 2023-12-31
tui: honour user's default visibility preference

lexiwinter:
Mastodon allows the user to configure a default visibility which should apply to all clients.  This setting is returned by the /api/v1/preferences method.

Fetch the user preferences when the TUI starts, and use it to set the default visibility when composing a new toot.  If the preference is not found, fall back to get_default_visibility().

### DIFF
--- a/toot/cli/tui.py
+++ b/toot/cli/tui.py
@@ -1,7 +1,7 @@
 import click
 
 from typing import Optional
-from toot.cli import TUI_COLORS, Context, cli, pass_context
+from toot.cli import TUI_COLORS, VISIBILITY_CHOICES, Context, cli, pass_context
 from toot.cli.validators import validate_tui_colors
 from toot.tui.app import TUI, TuiOptions
 
@@ -24,12 +24,18 @@ COLOR_OPTIONS = ", ".join(TUI_COLORS.keys())
     help=f"""Number of colors to use, one of {COLOR_OPTIONS}, defaults to 16 if
              using --color, and 1 if using --no-color."""
 )
+@click.option(
+    "-v", "--default-visibility",
+    type=click.Choice(VISIBILITY_CHOICES),
+    help="Default visibility when posting new toots; overrides the server-side preference"
+)
 @pass_context
 def tui(
     ctx: Context,
     colors: Optional[int],
     media_viewer: Optional[str],
     relative_datetimes: bool,
+    default_visibility: Optional[str]
 ):
     """Launches the toot terminal user interface"""
     if colors is None:
@@ -39,6 +45,7 @@ def tui(
         colors=colors,
         media_viewer=media_viewer,
         relative_datetimes=relative_datetimes,
+        default_visibility=default_visibility
     )
     tui = TUI.create(ctx.app, ctx.user, options)
     tui.run()
--- a/toot/api.py
+++ b/toot/api.py
@@ -618,6 +618,10 @@ def get_instance(base_url: str) -> Response:
     return http.anon_get(url)
 
 
+def get_preferences(app, user) -> Response:
+    return http.get(app, user, '/api/v1/preferences')
+
+
 def get_lists(app, user):
     return http.get(app, user, "/api/v1/lists").json()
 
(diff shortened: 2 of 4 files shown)


ihabunek:
Thanks, I was not familiar with the preferences API endpoint.

Looks mostly good, but visibility specified in toot settings should probably override the value fetched from preferences. 

What do you think about adding `--default-visiblity` option to `toot tui`? If the option is set, then we use that visibility, if not then fall back to the one in mastodon preferences.

Adding a CLI option would make it configurable via settings and env variables too.

lexiwinter:
what do you think about this?

i think this (still) means that `toot tui` and `toot post` might have a different default visibility if the user didn't configure it.  should `post` be updated to use the server preference as well?

toot/cli/tui.py
```diff
@@ -24,12 +24,18 @@
     help=f"""Number of colors to use, one of {COLOR_OPTIONS}, defaults to 16 if
              using --color, and 1 if using --no-color."""
 )
+@click.option(
+    "-V", "--default-visibility",
+    callback=validate_visibility,
```
ihabunek, line 29: You can avoid having a callback here:

```suggestion
    type=click.Choice(VISIBILITY_CHOICES),
```

lexiwinter:
fixed the test failure and updated the commit message.

toot/cli/tui.py
```diff
@@ -24,12 +24,18 @@
     help=f"""Number of colors to use, one of {COLOR_OPTIONS}, defaults to 16 if
              using --color, and 1 if using --no-color."""
 )
+@click.option(
+    "-V", "--default-visibility",
```
ihabunek, line 28: Any reason for having an uppercase `-V` instead of lowercase?

ihabunek (changes requested):

toot/cli/tui.py
lexiwinter: thanks, i changed that.

toot/cli/tui.py
lexiwinter: i used `-v` originally, but this is often used for a 'verbose' option by convention, so i went with `-V` instead to avoid a possible future conflict.  i don't have any particular preference though, i can switch it to `-v` if you like.

ihabunek:
> i think this (still) means that `toot tui` and `toot post` might have a different default visibility if the user didn't configure it. should `post` be updated to use the server preference as well?

This did occur to me, but reading preferences on each post would slow things down.

I need to test this, but I'm guessing posting a status without specifying the visibility would default to the one from preferences. If so, we could change the `--visibility` option to default to `null` instead of `public` and let the server set the visiblity according to user preferences.

toot/cli/tui.py
ihabunek: I doubt we'll have a verbose option for tui so please change it to `-v`. This will match the shorthand for `--visiblity` on `toot post`.

lexiwinter:
i tested and it worked like you expected:

```
diff --git a/toot/cli/post.py b/toot/cli/post.py
index afcfa4e..698a8d3 100644
--- a/toot/cli/post.py
+++ b/toot/cli/post.py
@@ -40,7 +40,7 @@ from toot.utils.datetime import parse_datetime
     "--visibility", "-v",
     help="Post visibility",
     type=click.Choice(VISIBILITY_CHOICES),
-    default="public",
+    default=None
 )
 @click.option(
     "--sensitive", "-s",
@@ -119,7 +119,7 @@ def post(
     media: Tuple[str],
     descriptions: Tuple[str],
     thumbnails: Tuple[str],
-    visibility: str,
+    visibility: Optional[str],
     sensitive: bool,
     spoiler_text: Optional[str],
     reply_to: Optional[str],
```

`toot post test` posted an unlisted toot (my server-side preference).

do you want me to include that in this PR or open a new one for it?

toot/cli/tui.py
lexiwinter: changed to -v

ihabunek (approved):

ihabunek:
Thanks!

Issue #469, opened 2024-02-03
Toot closes when a post's character limit is exceeded


thegreyshadow:
Hi there, thanks for toot. Seems a really nice application.

Using 0.41.1 in Slackware-current, 64-bit. When composing a toot (a Mastodon post), exceeding the server's character limits closes  the app and there's a message on the terminal (I assume it's stderr): 
```
Error: Validation failed: Text character limit of 500 exceeded
````
(in this case 500 was the limit of that particular server).

All text is lost, the application quits/crashes, and one has to open the TUI again

I think the app should have refused to post but not quitting and losing text. That is, I think it should handle this error more gracefully.

Thanks in advance,

Eduardo

PR #471, opened 2024-03-03
Fix for #469 compose shows error now when length too long

danschwarz:


ihabunek:
@danschwarz I don't want to do this because we don't always detect max length correctly, and I don't want the user not to be able to send a longer message just because fail at this.

Besides, calculating max length is more involved than just `len(content)`, beucause you need to take into account links which are shortened by the server.

If you look at https://mastodon.social/api/v1/instance you'll see `characters_reserved_per_url`:

```json
{
  "configuration": {
    "statuses": {
      "max_characters": 500,
      "max_media_attachments": 4,
      "characters_reserved_per_url": 23
    }
}
```

A better fix would be to catch the exception when posting and display the error to the user.

PR #472, opened 2024-03-03
EditBox now uses urwid_readline library. Fixes #353

### LINKED ISSUE #353: Feature Request: bash terminal shortcuts for text editing
Hey everyone!
Thanks for creating Toot, it's a beautiful piece of software and it's good to see it being actively developed!
I would love to be able to use bash terminal shortcuts while text editing, such as
```
Ctrl+A | move the cursor to the start of the line
Ctrl+E | move the cursor to the end of the line
Ctrl+U | clear the line towards the beginning
Ctrl+K | clear the line towards the end
Ctrl+W | delete last word
Ctrl+K | delete the rest of the line after the cursor
```

danschwarz:
The urwid_readline library allows readline style key commands. Equivalent to readline with set -o emacs. ctrl-a, ctrl-e, ctrl-k, etc.
The library doesn't have a mode equivalent to set -o vi, unfortunately, but this will add functionality sufficient to fix 353.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,8 @@ setup(
         "beautifulsoup4>=4.5.0,<5.0",
         "wcwidth>=0.1.7",
         "urwid>=2.0.0,<3.0",
-        "tomlkit>=0.10.0,<1.0"
+        "tomlkit>=0.10.0,<1.0",
+        "urwid_readline>=0.14"
     ],
     extras_require={
         # Required to display rich text in the TUI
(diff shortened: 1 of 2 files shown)


ihabunek:
Thanks, but I'm going to decline this one. Since `tooi` is coming along and aims to replace `toot tui`, I don't want to add new dependencies which are just used in the TUI, especially lesser known ones which are probably not packaged for various distros.

PR #473, opened 2024-03-09
Added safeguards to prevent crashes when rendering corrupt URLs

danschwarz:
Urwidgets' `Hyperlink` widget (correctly) throws an exception when presented with a URL containing disallowed characters (i.e. control characters.) If we get a corrupt URL in a status posting, this will cause a crash.  An example of such a status is [here](https://journa.host/@lolgop/112060018386619213). If this loads in the TUI today, ,it will crash.

With this PR, we will react to corrupt URLs by rendering as `urwid.Text` rather than `Hyperlink` and there will be no crash.

Given the recent comments about switching to tooi, this will probably be the last PR I submit for the toot tui.  It makes more sense to focus efforts on the tooi front end.

### DIFF
--- a/toot/tui/richtext/richtext.py
+++ b/toot/tui/richtext/richtext.py
@@ -60,7 +60,10 @@ def html_to_widgets(html, recovery_attempt=False) -> List[urwid.Widget]:
 
 
 def url_to_widget(url: str):
-    widget = len(url), urwid.Filler(Hyperlink(url, "link", url))
+    try:
+        widget = len(url), urwid.Filler(Hyperlink(url, "link", url))
+    except ValueError:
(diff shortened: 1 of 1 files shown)


marathone:
I'm just a simple user of this app, but I, for one, have appreciated all the recent work done by Mr. Schwarz! He was a breath of fresh air to what was a stale product before he arrived on the scene, in my opinion. No offence to ihabunek intended.

danschwarz:
@marathone thank you. I mean no offense to @ihabunek as well. As development is moving to a new front end, it makes sense to focus efforts there.

ihabunek:
Thanks!

PR #474, opened 2024-03-12
Fixes the KeyError for edited_at launch

skimave:
I saw an error preventing launching toot on my GoToSocial instance:

`File "/usr/lib/python3.11/site-packages/toot/tui/entities.py", line 56, in __init__
    if data["edited_at"]:
^^^^^^^^^^^^^^^^^
KeyError: 'edited_at'
An exception occurred, press X to view`

Seeing how `self.created_at` was being handled, I implemented the same kind of handling to `self.edited_at` including the original logic which fixed the launching issue at the same go.

### DIFF
--- a/toot/tui/entities.py
+++ b/toot/tui/entities.py
@@ -53,10 +53,7 @@ class Status:
         self.id = self.data["id"]
         self.account = self._get_account()
         self.created_at = parse_datetime(data["created_at"])
-        if data.get("edited_at"):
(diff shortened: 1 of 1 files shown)


ihabunek:
@skimave Thanks, but I believe that was already fixed in 4d5ac3cc4ebd392450f01f1020341a0209eadbaa.
Could you check if this happens for you in current master?

ihabunek:
\o/

PR #484, opened 2024-06-04
Improve Image Support

AnonymouX47:
- Fixes kitty graphics control sequences being displayed in iTerm2's title bar when image support is enabled.
- Optimizes image class selection.
- Optimizes image format pixel support detection.

The iTerm2 issue was brought to my attention by @danschwarz in https://github.com/AnonymouX47/term-image/issues/90#issuecomment-2144037780:

> Interestingly, the window title of the iTerm window has a part of a terminal command sequence in it, not sure what that's about.

The optimization changes were just things I noticed along the way.

### DIFF
--- a/toot/tui/images.py
+++ b/toot/tui/images.py
@@ -28,7 +28,6 @@ try:
         global _ImageCls
 
         if not _ImageCls:
-            BaseImage.forced_support = True
             _ImageCls = (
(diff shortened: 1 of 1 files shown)


danschwarz (approved):
Works well with iTerm now

PR #486, opened 2024-06-14
Added 'toot diag' command that outputs useful diagnostic info

danschwarz:
image

with `-f` flag, displays redacted versions of config and settings files

image


ihabunek:
Thanks!

PR #489, opened 2024-06-19
added _env to list of .flake8 excluded directories

danschwarz:


ihabunek:
This was added in another PR.

PR #492, opened 2024-07-10
Explain post visibility in toot post -h

strk:
I was missing the definitions in the help output, so I've added them


ihabunek:
Thanks.

PR #493, opened 2024-07-10
Allow running tests with pytest-3

strk:
On Ubuntu-22.04 installing python3-pytest installs pytest-3 and there is no pytest. This change allows running the first found on the system.


ihabunek:
Closing this one since I'd like to keep the makefile simple. This adds a ubuntu-specific check.

PR #500, opened 2024-08-28
CONTRIBUTING - Bug report - Replace toot env by toot diag

ArthurHoaro:
`toot diag` includes the same info as `toot env`, plus other things that can be useful in bug reporting.


ihabunek:
Thanks!

PR #512, opened 2024-12-08
Add alternate `redirect_uri` support for broken instances

rjp:
Pixelfed instances do not handle OOB OAuth correctly meaning you can't currently login to them using `toot` (e.g. https://github.com/pixelfed/pixelfed/issues/2522 )

This is a fudge to workaround that by allowing you to specify an alternate `redirect_uri` for broken servers which can take the HTTP redirect issued by Pixelfed and let you grab the code.

If you don't have a handy HTTP server, you can use `http://localhost` and your browser (at least Safari and Chrome) will have the code in the address bar for copying and pasting into `toot`.

I'm not 100% sure about how an empty `redirect_uri` is handled in `load_app` - it feels like supplying `-r` should set that instead of OOB on an app without an explicit setting but I don't think it's a situation that's going to crop up that often anyway.

(Apologies if this is a Pythonic mess, it's not one of my main languages)

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -45,6 +45,7 @@ def test_extract_active_user_app(sample_config):
     assert app.base_url == 'https://bar.social'
     assert app.client_id == 'ghi'
     assert app.client_secret == 'jkl'
+    assert app.redirect_uri == 'urn:ietf:wg:oauth:2.0:oob'
 
 
 def test_extract_active_when_no_active_user(sample_config):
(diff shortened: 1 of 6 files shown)


ihabunek:
Sorry I never got around to this one. On the brighter note, pixelfed has [fixed the issue](https://github.com/pixelfed/pixelfed/issues/2522#issuecomment-2570006487) and I successfully logged into my pixelfed account.

rjp:
> Sorry I never got around to this one.

No worries!  Sometimes these things just sort themselves out anyway (as this one did.)(

PR #513, opened 2024-12-09
Allow divergence in API

rakoo:
I'm using pleroma which doesn't have the same api:

- `/api/v1/preferences` doesn't exist
- `verified_at` doesn't always exist in a profile's field

Those changes make the absence of the first less verbose (instead of logging the exception inside the ui and mucking up the whole display, only the "Press X to view" is displayed) and the absence of the second one not crash

Thanks for this tool !


ihabunek:
Thanks!

PR #514, opened 2025-01-01
Handle urwid API deprecations

penguinolog:
* `_set_original_widget` -> `original_widget` property (setter)
* `button._state` -> `button.get_state()` (not deprecated, but we have public getter)


ihabunek:
Thanks!

PR #539, opened 2025-07-06
Fix: properly handle media_viewer commands with arguments

AbeEstrada:
This allows the `media_viewer` to be configured with both simple commands `feh` and commands with arguments `kitten icat`


ihabunek:
Thanks!

Issue #396, opened 2023-08-17
Timelines and read status


ffaf1:
When I browse my timeline (`toot tui`) I sometimes would like a “those toots are new” or “you have already read this” or “you have read until here ←” marker.

My workaround as now is to slap a ★ on the first message after each `toot tui` session, so I know what is new next time I invoke `toot tui`.

I wonder if I am using mastodon correctly — rather, whether I correctly understood mastodon timeline metaphor: maybe I should live with the flow and not aim for inbox zero, unlike what I do with a mail client or feed reader?

Consider a user question, and if the need is valid, a feature request.

danschwarz:
The Mastodon API has supported the concept of "markers" since version 3.0.0.  Basically these allow you to save and restore your position in the home or notifications timeline.  And this should work across Mastodon clients, as long as all the clients support the [marker API](https://docs.joinmastodon.org/methods/markers/).

`toot tui` does not yet support the marker API.  So this would be a feature request. Not sure what it would take to implement support, but I'll look into it when I have time.

Issue #222, opened 2022-11-11
TUI: Support "markers" API (preserve timeline read position)


sampsyo:
I would love for the TUI (and possibly also `toot timeline`?) to support [Mastodon's recent-ish `markers` API](https://docs.joinmastodon.org/methods/timelines/markers/), which saves and syncs a "last read" position between clients.

The exact UI approach is not entirely obvious. A simple approach would be to use the "cursor" position. That is, the TUI would select the last-read toot by default, and you would scroll up from there. Wherever you leave the cursor on exit, toot would save that as the last-read toot.

If something like this this seems palatable, I would be interested in working on it.

PR #546, opened 2025-09-03
Preserve focused status

AbeEstrada:
When refreshing the timeline, it now remembers (internally) the position of the last focused status. If the status isn’t there, it’ll pick the last one available at the bottom of the timeline.

I used the variable name `last_read_id` like in the [markers](https://docs.joinmastodon.org/methods/markers/) API, just in case it is implemented in the future.

Related: #222 #396


ihabunek:
Thanks!

PR #563, opened 2025-12-11
Fix broken tui.palette support

RobertKosten:
- Fixing what broke (I believe) when moving from `[tui]` to `[commands.tui]`, including the docs.
- Adding a `default` palette key that is applied to all unstyled content (via [urwid's `None` handling](http://urwid.org/manual/displayattributes.html#using-display-attributes))

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           uv sync
       - name: Run tests
         run: |
-          uv run pytest
+          uv run --all-extras pytest
       - name: Validate minimum required version
         run: |
-          uv run vermin --no-tips toot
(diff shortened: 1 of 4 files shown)


RobertKosten:
I've been digging into the test failure, since locally (python 3.11) `uv run pytest` (as run by github, AFAIK; Not too familiar with github actions or uv) works just fine:

It seems the problem lies with the optional `images` dependencies. In a fresh & simple checkout, `uv run pytest` fails, as it does here. In one where I've been running `uv run --all-extras toot-tui` a subsequent test run works fine, ergo the optional dependencies are probably the issue.

I'll add that flag to the test action and see what happens.

ihabunek:
Thanks!

PR #572, opened 2026-03-18
Fix backspace for wide chars in post

izkluxcvy:
When inputting wide characters (e.g. CJK, emoji) via input(), pressing backspace deleted only one byte at a time instead of one character, causing utf-8 decode error.

Fix by importing readline.


ihabunek:
Thanks!

PR #576, opened 2026-06-30
Fix table column alignment for wide characters

mahirhir:
## Summary

`print_table` (used by `toot lists`) sized its columns with `len(cell)` and padded them with `str.ljust(width)`. Both count code points, not terminal cells, so wide characters break the layout: CJK and emoji take two cells but were counted as one, which pushes every following column out of alignment.

The fix measures widths with `wcswidth()` and pads with `wcstring.pad()`, the wide-character aware helper already in the codebase. `wcswidth()` is also what `status_lines` uses to align the status header, so this just brings the table in line with the rest of the module. ASCII tables are unchanged.

## Before (a list whose title contains CJK)

```
ID  Title    Replies
--  -------  --------
1   Hello    list
2   日本語のリスト  followed
```

## After

```
ID  Title           Replies
--  --------------  --------
1   Hello           list
2   日本語のリスト  followed
```

## Tests

Added `tests/test_output.py`, which renders a table mixing ASCII and CJK rows and asserts every row ends up the same display width. It fails on the old code and passes with the fix. `pytest`, `flake8` and `vermin` all pass locally.

### DIFF
--- a/tests/test_output.py
+++ b/tests/test_output.py
@@ -1,0 +1,22 @@
+from wcwidth import wcswidth
+
+from toot.output import print_table
+
+
+def test_print_table_aligns_wide_characters(capsys):
+    # The Title column mixes ASCII and CJK entries. CJK characters occupy two
+    # terminal cells each, so column widths and padding must be measured in
+    # display columns rather than character count. When they are, every
+    # rendered row occupies the same total display width.
+    print_table(
+        ["ID", "Title"],
+        [
+            ["1", "Hello"],
+            ["2", "日本語"],
+        ],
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

ihabunek:
You're an impatient one. :) I merged your contribution: https://github.com/ihabunek/toot/commit/25b005770dfc7c49c16c4e346dc8b95c1e10afb4
Thanks, this is useful.

